PR #10, opened 2022-05-28
issue:3083189: ufm devices: add new parameter to reboot all devices

Abeer-Moghrabi:


haithamwj (changes requested):
why readme file does not include the new flag reboot_all?

PR #47, opened 2022-12-26
Adding Bright Integration plugin

ananalaghbar:


plugins/bright_plugin/src/bright-ui/src/app/bright/bright.component.html
```diff
@@ -0,0 +1,3 @@
+anan
```
haithamwj, line 1: what is this for? "does it print Anan?"

plugins/bright_plugin/src/bright-ui/src/app/bright/bright.routes.ts
```diff
@@ -0,0 +1,18 @@
...
+import {BrightComponent} from "./bright.component";
+import {DevicesJobsRoutes} from "./views/devices-jobs-view/devices-jobs.routes";
+import {DevicesJobsViewComponent} from "./views/devices-jobs-view/devices-jobs-view.component";
+
+export const BrightRoutes:Routes = [
+    {
+        path: 'bright',
+        component: BrightComponent,
+        children: [
+          // ...DevicesJobsRoutes
```
haithamwj, line 11: why this is commented out?

plugins/bright_plugin/src/bright-ui/src/app/bright/packages/bright/services/bright-backend.service.ts
```diff
@@ -0,0 +1,70 @@
...
+      const sign = filters.length ? '&' : '';
+      filters = filters + `${sign}from=${from}`;
+    }
+    if (to) {
+      const sign = filters.length ? '&' : '';
+      filters = filters + `${sign}to=${to}`;
+    }
+    if(from && to) {
+      filters = filters + `&tz=${this.getLocalTimezone()}`
+    }
```
haithamwj, line 62: filters are not clear to me, 
the '?' of the query string should not be depending on the first parameter 'from' and should appear if any parameter is provided, developer could pass to this function empty array of nodes and valid value for the parameter 'from'.
we could concatenate each query string parameter with it's own value, push it to array, then if the array is not empty, implode that array with '&&' and append '?' to the beginning of it

haithamwj: also why time zone applied only when we have both from and to ?

haithamwj: also why time zone applied only when we have both from and to ?

plugins/bright_plugin/src/bright-ui/src/app/bright/packages/time-picker-modal/time-picker-modal.constants.ts
```diff
@@ -0,0 +1,32 @@
+export class TimePickerModalConstants {
+    public static TIME_RANGES = [
+        {label: 'Last 5 Minutes', value: 'Last 5 Minutes', serverKey: '5'},
+        {label: 'Last 1 hour', value: 'Last 1 hour', serverKey: '60'},
+        {label: 'Last 12 hours', value: 'Last 12 hours', serverKey: '720'},
+        {label: 'Last 24 hours', value: 'Last 24 hours', serverKey: '1440'},
+        {label: 'Last week', value: 'Last week', serverKey: '10080'},
+        {label: 'Last month', value: 'Last month', serverKey: '43200'},
+        {label: 'Last 6 months', value: 'Last 6 months', serverKey: '259200'},
+        {label: 'Last 1 year', value: 'Last 1 year', serverKey: '525600'}
```
haithamwj, line 10: this is not last year, it's last 12 months! 
also the whole calculations of months and years can't be multiply of minutes, since months could be 28,29,30 or 31 days long 
but let me get back to you for this issue.

plugins/bright_plugin/src/bright-ui/src/app/constants/constants.ts
```diff
@@ -0,0 +1,13 @@
+import {Injectable} from '@angular/core';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class Constants {
+
+  public readonly baseApiUrl = "/ufmRestV2";
```
haithamwj, line 8: do we need to hard-code this in the plugin? I think we should leave it to the UFM

PR #59, opened 2023-02-14
3365803: Rename the current hello_world plugin to advanced_hello_world plugin

ananalaghbar:


plugins/advanced_hello_world_plugin/conf/advanced_hello_world_ui_conf.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "mfEntry": "/ufmRestV2/plugin/hello_world/files/remoteEntry.js",
+    "mfEntry": "/ufmRestV2/plugin/advanced_hello_world/files/remoteEntry.js",
```
haithamwj, line 3: we need relative path

PR #67, opened 2023-02-16
Harden Slurm integration scripts by default

flx42:


flx42:
Added a new commit to also stop using `SLURM_STEP_ID` as it was not being used and it is not available in `SlurmctldProlog` anyway

anasbadaha (approved):

plugins/SLURM-Integration/ufm-prolog.sh
```diff
@@ -13,4 +13,4 @@
 echo "==Pre job==:"
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 cd $SCRIPTPATH
-python3 ufm_slurm_prolog.py -i $SLURM_JOBID -s $SLURM_STEP_ID || true
+python3 ufm_slurm_prolog.py -i $SLURM_JOBID || true
```
mike-dubman: why remove STEP_ID? it is documented and exists in latest SLURM (AFAIS)

flx42: I explained here: https://github.com/Mellanox/ufm_sdk_3.0/pull/67#issuecomment-1433699972

1. It was not being used in the code.
2. It is not available in `PrologSlurmctld` or `EpilogSlurmctld` where those scripts are supposed to be called from. See https://slurm.schedmd.com/prolog_epilog.html: `SLURM_STEPID Step ID of the current job. Available in SrunProlog and SrunEpilog.`

mike-dubman: But isn’t same script called from srun prolog/epilog?
@anasbadaha - why it is not used?

flx42: No, this script isn't designed to run in srun prolog/epilog (usually unprivileged).

mike-dubman: according to doc https://slurm.schedmd.com/prolog_epilog.html prolog/epilog is called for job step alloc/dealloc

anasbadaha: @mike-dubman As I understood from Ibrahim Baraka who added the Use of this param "STEP_ID" that by removing this param the UFM-SLURM plugin will fail incase STEP_ID was used in SLURM_JOB. I think we need to wait and test the code with and without STEP_ID before removing it.

flx42:
Rebased my branch on the latest `main` branch, I realized my working branch was a bit behind.

flx42:
ping @anasbadaha?

kedeme:
Confirming that we are not using the SLURM step id in UFM-SLURM integration.

PR #88, opened 2023-04-19
issue:3439575 Manage topoconfig file creation and NDT file status and

alextabachnik:
capabilities

Add API to change boundary port state and deploy at to UFM at the same request


kedeme (changes requested):
please fix typo on line #153:
onve run verification once - status become verified. 
-->
once verification passed - status become verified.

PR #92, opened 2023-04-25
issue:3445890 Add port regular expression for BF port

alextabachnik:
for port number/name like "Port B4D0F0/0/0"


alextabachnik:
Regular expression update. No functionality changes.

PR #93, opened 2023-04-27
issue:3290042 added support for dynamic telemetry

drorlevy:


plugins/pdr_deterministic_plugin/ufm_sim_web_service/isolation_mgr.py
```diff
@@ -350,15 +377,33 @@ def get_isolation_state(self):
                 port_state.update(Constants.STATE_ISOLATED, Constants.ISSUE_OONOC)
                 self.ports_states[port] = port_state
 
+    def start_telemetry_session(self):
+        self.logger.info("Starting telemetry session")
+        response = self.ufm_client.start_dynamic_session(Constants.PDR_DYNAMIC_NAME, self.telemetry_counters, self.t_isolate)
+        if response and response.status_code == http.HTTPStatus.ACCEPTED:
+            port = str(response.content)
+        else:
+            self.logger.error(f"Failed to start dynamic session: {response}")
```
kedeme, line 386: what happens in case of start session failed - will the plugin retry to start the session later on?

plugins/pdr_deterministic_plugin/ufm_sim_web_service/isolation_mgr.py
```diff
@@ -350,15 +377,33 @@ def get_isolation_state(self):
...
+            return False
+        return port       
+
+    def run_telemetry_get_port(self):
+        while not self.ufm_client.running_dynamic_session(Constants.PDR_DYNAMIC_NAME):
+            endpoint_port = self.start_telemetry_session()
+            time.sleep(20)
```
kedeme, line 393: timeout (20 seconds) should be configurable - depending on the fabric size...

kedeme (changes requested):

plugins/pdr_deterministic_plugin/ufm_sim_web_service/isolation_mgr.py
```diff
@@ -255,7 +276,11 @@ def read_next_set_of_high_ber_or_pdr_ports(self):
                         port_data[Constants.ASIC] = port_asic
                         port_data[Constants.WIDTH] = port_width
                     if not port_asic:
-                        logger.debug(f"Couldn't retrieve HW technology for port {port_name}, can't verify it's BER values")
+                        self.logger.debug(f"Couldn't retrieve HW technology for port {port_name}, can't verify it's BER values")
```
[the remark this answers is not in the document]
drorlevy: it can be thousands of lines each time that we retrieve telemetry

plugins/pdr_deterministic_plugin/ufm_sim_web_service/isolation_mgr.py
drorlevy: it is not related to the discovery but to a sleep in the dynamic

kedeme (approved):

PR #104, opened 2023-05-26
Refuse to proceed if ufm_slurm.conf is world-readable

flx42:
This file contains a password or an API token in plaintext, the administrator should be warned that this is unsafe.

Also change mode for ufm_slurm.conf to remove execute permissions.


plugins/SLURM-Integration/ufm_slurm.conf
anasbadaha, line 1: why this is an empty file?

plugins/SLURM-Integration/ufm_slurm_utils.py
```diff
@@ -99,6 +100,10 @@ def getSlurmConfFile(self):
 
     def read_conf_file(self, key):
         conf_file = self.getSlurmConfFile()
+        conf_mode = os.stat(conf_file).st_mode
+        if conf_mode & 0o77 != 0:
+            logging.error(f"Permissions for configuration file {conf_file} are too open: {oct(conf_mode & 0o777)}")
```
anasbadaha, line 105: could you please tell why do you need this change?

anasbadaha (changes requested):

plugins/SLURM-Integration/ufm_slurm.conf
flx42: It's not an empty file, the file changed mode.

plugins/SLURM-Integration/ufm_slurm_utils.py
flx42: Please take a look at the comment in the patch:
> This file contains a password or an API token in plaintext, the administrator should be warned that this is unsafe.

Does that make sense?

It's currently too easy to setup those scripts and end up with `ufm_slurm.conf` that is readable by all cluster users, which mean they would now have access to the whole UFM REST API.

kedeme (approved):

anasbadaha:
Hi @flx42,
We wanted to inform you that we have made a decision not to perform permission checking for the ufm_slurm.conf file inside the read_conf_file function. This function is called multiple times for each submitted job, and instead we want to ensure that the ufm_slurm.conf file is installed with the default permission of 644.

flx42:
Why does it need to be world readable? That's a security risk.

PR #110, opened 2023-06-26
Bright CI

mohfh:


plugins/bright_plugin/.ci/ci_matrix.yaml
```diff
@@ -0,0 +1,54 @@
...
+kubernetes:
+  cloud: swx-k8s-spray
+
+volumes:
+  - {mountPath: /var/run/docker.sock, hostPath: /var/run/docker.sock}
+  - {mountPath: /auto/UFM, hostPath: /auto/UFM }
+
+
+runs_on_dockers:
+   - {file: '.ci/Dockerfile', arch: 'x86_64', name: 'plugin_worker', tag: 'latest'}
```
lennybe, line 22: if this docker is based on specific Dockerfile, please set a name of it related to this plugin,
otherwise this docker container can be overwritten on harbor or can be deleted manually since it's unclear where it belongs.

plugins/bright_plugin/.ci/ci_matrix.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+
+steps:
+  - name: Build Plugin
+    containerSelector: "{name: 'plugin_worker'}"
+    run: |
+      cd plugins/$plugin_dir/build
+      bash -x ./docker_build.sh latest /
+      ls -l /
```
lennybe, line 31: pls save return value and exit on error.
at this point this step always returns 0 and never fails

lennybe (changes requested):

PR #112, opened 2023-07-10
Issue: 3482083 Change the permission of ufm_slurm.conf to be 600 by

anasbadaha:
default
Issue: 	3534518: UFM-SLURM integration script destroys slurm functionality

### DIFF
--- a/plugins/SLURM-Integration/install.sh
+++ b/plugins/SLURM-Integration/install.sh
@@ -64,6 +64,19 @@ function update_slurm_conf()
     fi
 }
 
+function update_python_path_for_ufm_prolog_epilog()
+{
+    python_path=$(which python3)
+    sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_PROLOG_FILE
+    if [[ $? -ne 0 ]];then
+        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in ${SLURM_DIR/$UFM_PROLOG_FILE} file, please try to do it manually."
+    fi
+    sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_EPILOG_FILE
+    if [[ $? -ne 0 ]];then
+        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in ${SLURM_DIR/$UFM_EPILOG_FILE} file, please try to do it manually."
+    fi
+}
+
 function validate_requirements()
 {
     # Check if script is running on SLURM controller
@@ -246,6 +259,7 @@ copy_integration_files
 echo $CONF_SETTINGS
 update_slurm_conf $PROLOG_SLURMCTLD "$SLURM_DIR/$UFM_PROLOG_FILE"
 update_slurm_conf $EPILOG_SLURMCTLD "$SLURM_DIR/$UFM_EPILOG_FILE"
+update_python_path_for_ufm_prolog_epilog
 echo $INSTALLATION_COMPLETED_SUCCESSFULLY
 echo $SET_UFM_SLURM_CONF
 exit 0
--- a/plugins/SLURM-Integration/install.sh
+++ b/plugins/SLURM-Integration/install.sh
@@ -67,7 +67,7 @@ function update_slurm_conf()
 function set_python3_path()
 {   local full_python3_path=$1
     local file_name=$2
-    sed -i -e "s#python3_path#$full_python3_path#g" $SLURM_DIR/$file_name
+    sudo sed -i -e "s#python3_path#$full_python3_path#g" $SLURM_DIR/$file_name
     if [[ $? -ne 0 ]];then
         echo "Failed to set PYTHONPATH for $file_name, please set it manually at $SLURM_DIR/$file_name by setting python3_path parameter, prior to running the UFM-Slurm integration."
     fi
@@ -80,6 +80,7 @@ function update_python_path_for_ufm_prolog_epilog()
     python3_path=$(type -p python3)
     if [[ $? -ne 0 ]];then
         echo "Failed to run (type -p python3) command, please make sure that python3 is installed and configured appropriately, prior to running the UFM-Slurm integration."
+        exit 2
     fi
     python3_path=$(echo "$python3_path" | cut -d ' ' -f 3)
     set_python3_path $python3_path $UFM_PROLOG_FILE
--- a/plugins/SLURM-Integration/install.sh
+++ b/plugins/SLURM-Integration/install.sh
@@ -69,11 +69,11 @@ function update_python_path_for_ufm_prolog_epilog()
     python_path=$(which python3)
     sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_PROLOG_FILE
     if [[ $? -ne 0 ]];then
-        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in ${SLURM_DIR/$UFM_PROLOG_FILE} file, please try to do it manually."
+        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in $SLURM_DIR/$UFM_PROLOG_FILE file, please try to do it manually."
     fi
     sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_EPILOG_FILE
     if [[ $? -ne 0 ]];then
-        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in ${SLURM_DIR/$UFM_EPILOG_FILE} file, please try to do it manually."
+        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in $SLURM_DIR/$UFM_EPILOG_FILE file, please try to do it manually."
     fi
 }
 
--- a/plugins/SLURM-Integration/install.sh
+++ b/plugins/SLURM-Integration/install.sh
@@ -64,17 +64,26 @@ function update_slurm_conf()
     fi
 }
 
+function set_python3_path()
+{   local full_python3_path=$1
+    local file_name=$2
+    sed -i -e "s#python3_path#$full_python3_path#g" $SLURM_DIR/$file_name
+    if [[ $? -ne 0 ]];then
+        echo "Failed to set PYTHONPATH for $file_name, please set it manually at $SLURM_DIR/$file_name by setting python3_path parameter, prior to running the UFM-Slurm integration."
+    fi
+}
+
 function update_python_path_for_ufm_prolog_epilog()
+# This function is used to get the full path of python3 and replace it with keyward python3_path
+# In both ufm-prolog and ufm-epilog files
 {
-    python_path=$(which python3)
-    sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_PROLOG_FILE
+    python3_path=$(type -p python3)
     if [[ $? -ne 0 ]];then
-        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in $SLURM_DIR/$UFM_PROLOG_FILE file, please try to do it manually."
+        echo "Failed to run (type -p python3) command, please make sure that python3 is installed and configured appropriately, prior to running the UFM-Slurm integration."
     fi
-    sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_EPILOG_FILE
-    if [[ $? -ne 0 ]];then
-        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in $SLURM_DIR/$UFM_EPILOG_FILE file, please try to do it manually."
-    fi
+    python3_path=$(echo "$python3_path" | cut -d ' ' -f 3)
+    set_python3_path $python3_path $UFM_PROLOG_FILE
+    set_python3_path $python3_path $UFM_EPILOG_FILE
 }
 
 function validate_requirements()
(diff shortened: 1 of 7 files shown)


plugins/SLURM-Integration/install.sh
```diff
@@ -64,6 +64,19 @@ function update_slurm_conf()
     fi
 }
 
+function update_python_path_for_ufm_prolog_epilog()
```
kedeme, line 67: add function description

plugins/SLURM-Integration/install.sh
```diff
@@ -64,6 +64,19 @@ function update_slurm_conf()
...
+{
+    python_path=$(which python3)
+    sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_PROLOG_FILE
```
kedeme, line 70: lines 70-73 export to function: set_python3_path(path)
then call this function from here:
set_python3_path($SLURM_DIR/$UFM_PROLOG_FILE)
set_python3_path($SLURM_DIR/$UFM_EPILOG_FILE)

plugins/SLURM-Integration/install.sh
```diff
@@ -64,6 +64,19 @@ function update_slurm_conf()
...
+    if [[ $? -ne 0 ]];then
+        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in $SLURM_DIR/$UFM_PROLOG_FILE file, please try to do it manually."
```
kedeme, line 72: Failed to set PYTHONPATH for $filename (prolog / epilog), please set it manually at $filepath by setting "python3_path" parameter, prior to running the UFM-Slurm integration.

plugins/SLURM-Integration/ufm_slurm.conf
```diff
@@ -19,4 +19,6 @@ sharp_allocation=false
 # By setting partially_alloc to false, UFM SHARP allocation API will fail the SHArP allocation request if at least one node does not exist in the fabric.
 partially_alloc=true
 # Name of UFM-SLURM integration logging file
-log_file_name=ufm_slurm.log
\ No newline at end of file
+log_file_name=ufm_slurm.log
+# By setting this parameter to a non-zero value, the running Slurm job will terminate if either ufm-prolog.sh or ufm-epilog.sh exits with a non-zero status.
```
kedeme, line 23: please add:
by default, the value is set to 0, which means, allowing Slurm job to continue running even if the ufm-prolog.sh or ufm-epilog.sh exit with non-zero status.

plugins/SLURM-Integration/ufm_slurm.conf
```diff
@@ -19,4 +19,6 @@ sharp_allocation=false
...
+should_fail_slurm_job=0
```
kedeme, line 24: should_fail_slurm_job --> fail_slurm_job_upon_failure

plugins/SLURM-Integration/ufm_slurm_base.py
```diff
@@ -48,7 +48,7 @@ def init(self):
         self.app_resources_limit = self.general_utils.get_conf_parameter_value(Constants.CONF_APP_RESOURCES_LIMIT)
         if self.app_resources_limit and int(self.app_resources_limit) < -1:
             logging.error("app_resources_limit param must be an integer number greater than -1!")
-            sys.exit(1)
+            sys.exit(self.should_fail)
```
kedeme, line 51: why to fail prolog and epilog because of wrong configuration value?
in case wrong configuration value was set (-2) set the default value and notify the user with message:
logging.error("app_resources_limit param must be an integer number greater than -1, (got $config_resource_limit), using default value - 1")

Then no need to exit the script for that and in that case we can remove the "should_fail_slurm_job" from configuration + remove all related logic

kedeme (changes requested):

kedeme (changes requested):
Also, I could not find the change of configuration permission to 600 in this PR!
can you please check?

plugins/SLURM-Integration/install.sh
mike-dubman, line 72: typo: s/pythn3/python3

plugins/SLURM-Integration/install.sh
```diff
@@ -64,6 +64,19 @@ function update_slurm_conf()
...
+    fi
+    sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_EPILOG_FILE
+    if [[ $? -ne 0 ]];then
+        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in $SLURM_DIR/$UFM_EPILOG_FILE file, please try to do it manually."
```
mike-dubman, line 76: typo

plugins/SLURM-Integration/ufm-epilog.sh
```diff
@@ -12,4 +12,4 @@
 #
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 cd $SCRIPTPATH
-python3 ufm_slurm_epilog.py -i $SLURM_JOBID
+sudo python3_path ufm_slurm_epilog.py -i $SLURM_JOBID
```
mike-dubman, line 15: why need sudo? AFAIR prolog/epilog is executed as root

plugins/SLURM-Integration/install.sh
```diff
@@ -64,6 +64,19 @@ function update_slurm_conf()
     fi
 }
 
+function update_python_path_for_ufm_prolog_epilog()
+{
+    python_path=$(which python3)
```
mike-dubman, line 69: 1. use ```type python3```, as ```which``` command is external and not mandatory. ```type``` command is shell built-in command which is always available in any distro
2. what if ```which python3``` fails to find python3? dont see proper error flow.
in line 71-72 the check and err message are wrong:

need to check that ```which python3``` succeeded to run
if succeeded to run - check that command result is executable file
replace with sed and only then give error that failed to replace (also check that conf file is writable)
think of all what can go wrong and help user to navigate

plugins/SLURM-Integration/install.sh
```diff
@@ -64,6 +64,19 @@ function update_slurm_conf()
...
+    sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_PROLOG_FILE
+    if [[ $? -ne 0 ]];then
+        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in $SLURM_DIR/$UFM_PROLOG_FILE file, please try to do it manually."
+    fi
+    sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_EPILOG_FILE
```
mike-dubman, line 74: same, check if really need sudo and granular error check with user facing messages

plugins/SLURM-Integration/ufm_slurm_base.py
anasbadaha: I fixed the logging and not exiting the code in case of invalid value, regarding the removal of should_fail_slurm_job we can't remove it because it's used in many pleases.

anasbadaha:
> Also, I could not find the change of configuration permission to 600 in this PR! can you please check?

I set the file permission to be 600 by default, and on installation will install it with default permission and not change it.

plugins/SLURM-Integration/ufm-epilog.sh
anasbadaha: fixed

plugins/SLURM-Integration/install.sh
anasbadaha: I checked it and we need to run it with sudo.

mike-dubman: maybe run entire script with ```sudo install.sh``` and not specific lines?
sudo is not default everywhere
script can also check that active user is root and fail if not.

plugins/SLURM-Integration/install.sh
```diff
@@ -64,17 +64,26 @@ function update_slurm_conf()
...
+}
+
+function update_python_path_for_ufm_prolog_epilog()
+# This function is used to get the full path of python3 and replace it with keyward python3_path
+# In both ufm-prolog and ufm-epilog files
+{
+    python3_path=$(type -p python3)
     if [[ $? -ne 0 ]];then
-        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in $SLURM_DIR/$UFM_EPILOG_FILE file, please try to do it manually."
+        echo "Failed to run (type -p python3) command, please make sure that python3 is installed and configured appropriately, prior to running the UFM-Slurm integration."
```
mike-dubman, line 82: where ```exit $errorcode``` happens when taking error flow?

kedeme (changes requested):

plugins/SLURM-Integration/install.sh
anasbadaha: I'm exiting the installation in case I failed to run  type -p python3 command.

plugins/SLURM-Integration/install.sh
anasbadaha: I tried to run the entire install script with sudo as well, but it failed with permission error to the sed command "sed: couldn't edit /etc/slurm/: not a regular file", farther that we are running all the command inside the install script with sudo and if I will change it now I'm sure that we will have many issues with permissions.

plugins/SLURM-Integration/ufm_slurm_base.py
```diff
@@ -47,8 +47,9 @@ def init(self):
         self.partially_alloc = self._toBoolean(self.partially_alloc, Constants.CONF_PARTIALLY_ALLOC, True)
         self.app_resources_limit = self.general_utils.get_conf_parameter_value(Constants.CONF_APP_RESOURCES_LIMIT)
         if self.app_resources_limit and int(self.app_resources_limit) < -1:
-            logging.error("app_resources_limit param must be an integer number greater than -1!")
-            sys.exit(1)
+            logging.error(
+                "app_resources_limit param must be an integer number greater than -1, (got {0}), using default value - 1".format(self.app_resources_limit))
```
kedeme, line 51: typo:
using default value: -1

plugins/SLURM-Integration/install.sh
```diff
@@ -67,7 +67,7 @@ function update_slurm_conf()
 function set_python3_path()
 {   local full_python3_path=$1
     local file_name=$2
-    sed -i -e "s#python3_path#$full_python3_path#g" $SLURM_DIR/$file_name
+    sudo sed -i -e "s#python3_path#$full_python3_path#g" $SLURM_DIR/$file_name
```
mike-dubman, line 70: why no exit with error code after line 72? or it is in diff place?

anasbadaha: we don't want to exit the installation in case of failure, because we are notifying the user about the failure about the failure and we are asking him to do it manually. you can fine it the error message in line 72.

plugins/SLURM-Integration/install.sh
```diff
@@ -64,17 +64,26 @@ function update_slurm_conf()
...
-{
-    python_path=$(which python3)
-    sudo sed -i -e "s#python3_path#$python_path#g" $SLURM_DIR/$UFM_PROLOG_FILE
+function set_python3_path()
+{   local full_python3_path=$1
+    local file_name=$2
+    sed -i -e "s#python3_path#$full_python3_path#g" $SLURM_DIR/$file_name
     if [[ $? -ne 0 ]];then
-        echo "Warning:: Failed to replace the output of 'which pythn3' command with python3_path keyword in $SLURM_DIR/$UFM_PROLOG_FILE file, please try to do it manually."
+        echo "Failed to set PYTHONPATH for $file_name, please set it manually at $SLURM_DIR/$file_name by setting python3_path parameter, prior to running the UFM-Slurm integration."
```
mike-dubman, line 72: exit with error?

plugins/SLURM-Integration/install.sh
mike-dubman: where? dont see it

plugins/SLURM-Integration/install.sh
anasbadaha: it will just print the error message to the user, I don't think we need to exit the installation because the user can do it manually in case it's failed to update the files, what do you think?

mike-dubman: we should always abort on error with clear err

plugins/SLURM-Integration/install.sh
mike-dubman: we should exit on 1st error and not continue, so user will see error, fix and re-run
also - why sudo is inside script and not outside
also need to check that $file_name is writeable and fail if not
check if token #python3_path# exists in file and do sed only when it does and provide clear error what is wrong

anasbadaha: working on the fixes, regarding the sudo I tried to remove it from the script and run sudo ./install.sh but we faced many issues with permission, so decided to keep the sudo inside the script.

kedeme (approved):

plugins/SLURM-Integration/install.sh
mike-dubman: what issues? using sudo inside the script is not a good approach, as not every slurm install allows/has sudo and admin will need to change script and remove sudo as cmd

plugins/SLURM-Integration/install.sh
mike-dubman: ```couldn't edit /etc/slurm/:``` because actual filename is missing in the string. seems bug in a code

lennybe:
@kedeme , 
why did you merge code that failed on CI?

lennybe:
@anasbadaha , please add build step 
http://hpc-master.lab.mtl.com:8080/blue/organizations/jenkins/ufm_sdk_3.0/detail/ufm_sdk_3.0/369/pipeline


++ uniq

+ plugin_dir_name=plugins/SLURM-Integration

+ '[' -f ../plugins/SLURM-Integration/.ci/ci_matrix.yaml ']'

+ echo 'Error: CI configuration file for plugins/SLURM-Integration not found.'

Error: CI configuration file for plugins/SLURM-Integration not found.

+ exit 1

PR #115, opened 2023-07-16
issue:3538330 Avoid writing to topoconfig file record for miswired links

alextabachnik:
In case and for peer in NDT file set incorrect port number in link between switches - the link defined as miswired and should not appear in topoconfig file with incorrect peer GUID, as it will fail opensm.

### DIFF
--- a/plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
+++ b/plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
@@ -549,7 +549,7 @@ def create_topoconfig_file(links_info_dict, ndt_file_path, patterns,
                     error_message = "Failed to detect peer port GUID. Skip link in topoconfig file: peer device: {} peer port: {}".format(peer_device, peer_port)
                     logging.error(error_message)
                     file_creation_failed = True
-                    failed_lable_conversion.append(port_key.replace("___", " Port:"))
+                    failed_lable_conversion.append(link_key_peer.replace("___", " Port:"))
                     continue
                 # in case peer port is not a number - get number from mapping
                 if not peer_port.isnumeric():
--- a/plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
+++ b/plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
@@ -507,7 +507,6 @@ def create_topoconfig_file(links_info_dict, ndt_file_path, patterns,
                 peer_device, peer_port, error_message = parse_ndt_port(
                                          os.path.basename(ndt_file_path), row, index,
                                          PortType.DESTINATION, patterns, True)
-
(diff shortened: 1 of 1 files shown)


plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
```diff
@@ -543,6 +542,15 @@ def create_topoconfig_file(links_info_dict, ndt_file_path, patterns,
...
                 peer_port_guid = links_info_dict.get(link_key_peer)
+                if not peer_port_guid:
+                    # unwired link: incorrect peer port number. Such link should
+                    # not be written into topoconfig file - if will cause opensm malfunction
+                    # such line should not appear in topoconfig - continue
+                    error_message = "Failed to detect peer port GUID. Skip link in topoconfig file: peer device: {} peer port: {}".format(peer_device, peer_port)
+                    logging.error(error_message)
+                    file_creation_failed = True
+                    failed_lable_conversion.append(port_key.replace("___", " Port:"))
+                    continue
```
atolikin94, line 553: are you sure we want to continue in case of such link? doesn't it make sense to exit and report an error? because I suppose the whole NDT file should be correct, and if there is an error - user should fix it before proceeding..

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
```diff
@@ -543,6 +542,15 @@ def create_topoconfig_file(links_info_dict, ndt_file_path, patterns,
             if peer_device and peer_port:
                 link_key_peer = "%s___%s" % (peer_device, peer_port)
                 peer_port_guid = links_info_dict.get(link_key_peer)
+                if not peer_port_guid:
+                    # unwired link: incorrect peer port number. Such link should
+                    # not be written into topoconfig file - if will cause opensm malfunction
+                    # such line should not appear in topoconfig - continue
+                    error_message = "Failed to detect peer port GUID. Skip link in topoconfig file: peer device: {} peer port: {}".format(peer_device, peer_port)
+                    logging.error(error_message)
+                    file_creation_failed = True
```
atolikin94, line 551: is this variable used somewhere?

kedeme (commented):
on the next PR for NDT, please convert all string formatting to f-string

lennybe:
bot:retest

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
alextabachnik: according to architecture we should allow used to deploy topoconfig even if there are problems during verification.
so to avoid opensm failure decided not to include links from NDT with problems in topoconfig file. UI will send notification that it is on user responsibility to deploy topoconfig when verification failed

kedeme:
please add regression test for UFM to cover the miswired use case.

PR #118, opened 2023-07-19
issue:3518592 Remove reports generated for NDT if NDT file removed

alextabachnik:
Removing report generated for NDT file verification if removed NDT file itself


plugins/UFM_NDT_Plugin/ufm_sim_web_service/merger_resources.py
```diff
@@ -381,6 +382,32 @@ def __init__(self):
...
+        :param ndt: name of NDT file
+        '''
+        # find reports related to NDT file name and remove them.
+        # Update reports.json file
+        try:
+            with open(self.reports_list_file, "r", encoding="utf-8") as reports_list_file:
+                # unhandled exception in case reports file was changed manually
+                data = json.load(reports_list_file)
+                data_to_proceed = data.copy()
+                for report_data in data_to_proceed:
```
shmufeld, line 397: this loop can be outside the "with" statement?

atolikin94 (approved):
Don't really understand why do we need to remove reports when NDT was remover, but the code looks good.

plugins/UFM_NDT_Plugin/ufm_sim_web_service/merger_resources.py
alextabachnik: Possible to open file for read and  read data and then to close. 
With doing the same. In this case both options are valid.

shmufeld: you right. it's possible but I thought it will be cleaner...never mind.

shmufeld (approved):

kedeme (approved):

kedeme (approved):

PR #120, opened 2023-08-07
issue: Normalize length of port guid for port number search

alextabachnik:
in ibdiagnet2.net_dump file port GUID does not contain prefix zero and in ibdiagnet2.db_csv - contains. So failed to get port number based on specific port GUID as it does not match:
in
/tmp/ndt_plugin/ibdiagnet2.net_dump
0xc42a103008b3b49
in
/tmp/ndt_plugin/ibdiagnet2.db_csv
0x0c42a103008b3b49

need to normalize port GUID and to add prefix 0 where needed.

### DIFF
--- a/plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
+++ b/plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
@@ -11,15 +11,16 @@
 # This software product is governed by the End User License Agreement
 # provided with the software product.
 #
-import os
+import csv
+from datetime import datetime
+import json
 import logging
+import os
+import re
 import subprocess
 import time
-import json
-from datetime import datetime
-import csv
+
 import pandas as pd
-import re
 from topo_diff.topo_diff import parse_ndt_port, PortType, parse_ibdiagnet_dump
 
 
@@ -168,6 +169,16 @@ def run_ibdiagnet_verification_command():
     '''
     status, cmd_output = execute_generic_command(IBDIAGNET_PORT_VERIFICATION_COMMAND)
     return status
+
+def normalize_port_guid_lengh(port_guid):
+    '''
+    Normalize port guid len - in need add prefix 0
+    :param port_guid:
+    '''
+    port_guid_len = len(port_guid)
+    if port_guid_len < 18: # in case and we have guid that is short - need to add zero
+        port_guid = "".join(["0x", "%s"%("0"*(18-port_guid_len)), port_guid[2:]])
+    return port_guid
 
 def get_switch_port_label2port_num_map():
     '''
@@ -527,7 +538,11 @@ def create_topoconfig_file(links_info_dict, ndt_file_path, patterns,
             if not start_device in device_to_guid_map:
                 device_to_guid_map[start_device] = port_guid
             if not start_port.isnumeric():
-                port_key = "%s___%s" % (port_guid, start_port)
+                # port guid in this case could have wrong lenght, as some reason
+                # the guid in net_dump file does not have prefix 0, but in db_csv - has
+                # so need to normalize
+                key_port_guid = normalize_port_guid_lengh(port_guid)
+                port_key = "%s___%s" % (key_port_guid, start_port)
                 start_port_num = node_guid_lable2port_num.get(port_key, None)
                 if start_port_num:
                     start_port = str(start_port_num)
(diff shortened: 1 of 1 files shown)


kedeme (approved):

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
```diff
@@ -169,6 +170,16 @@ def run_ibdiagnet_verification_command():
     status, cmd_output = execute_generic_command(IBDIAGNET_PORT_VERIFICATION_COMMAND)
     return status
 
+def normalize_port_guid_length(port_guid):
+    '''
+    Normalize port guid len - in need add prefix 0
+    :param port_guid:
+    '''
+    port_guid_len = len(port_guid)
+    if port_guid_len < 18: # in case and we have guid that is short - need to add zero
```
kedeme, line 179: please add debug log message for this condition..
please add automation test / unit test to cover this use case.

kedeme (changes requested):
please add automation / unit-test to cover this use case.

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
alextabachnik: 1. No need debug message for such flow - it will be a lot records in log with no any reason.
2. In general tests written by Anas checking this issue on verification setup. Normal flow should work on verification setup. Otherwise we will see errors in log file and incorrect topoconfig file.

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
```diff
@@ -169,6 +170,16 @@ def run_ibdiagnet_verification_command():
     status, cmd_output = execute_generic_command(IBDIAGNET_PORT_VERIFICATION_COMMAND)
     return status
 
+def normalize_port_guid_length(port_guid):
+    '''
+    Normalize port guid len - in need add prefix 0
```
shmufeld, line 175:
```suggestion
    Normalize port guid len - if need add prefix 0
```

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
```diff
@@ -169,6 +170,16 @@ def run_ibdiagnet_verification_command():
...
+    :param port_guid:
+    '''
+    port_guid_len = len(port_guid)
+    if port_guid_len < 18: # in case and we have guid that is short - need to add zero
+        port_guid = "".join(["0x", "%s"%("0"*(18-port_guid_len)), port_guid[2:]])
```
shmufeld, line 180: f strings
```suggestion
        port_guid = "".join(["0x", f"{"0" * (18-port_guid_len)}"), port_guid[2:]])
```

shmufeld (changes requested):

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
atolikin94, line 179: why are we ignoring guids longer than 18 symbols?
isn't it a problem that UFM needs to report and fail on it? or at least maybe log that guid

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
alextabachnik: It is the same. And it is additional location of such command - so to be aligned to the code.

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
alextabachnik: guid is a number and it could be 0x887a or 0x000887a
the problem is that in ibdiagnet.net_dump and in ibdiagnet.db_csv the format is not the same ... so need to make them of the same length (max between them).

plugins/UFM_NDT_Plugin/ufm_sim_web_service/topo_diff/ndt_infra.py
```diff
@@ -527,7 +538,11 @@ def create_topoconfig_file(links_info_dict, ndt_file_path, patterns,
             if not start_device in device_to_guid_map:
                 device_to_guid_map[start_device] = port_guid
             if not start_port.isnumeric():
-                port_key = "%s___%s" % (port_guid, start_port)
+                # port guid in this case could have wrong lenght, as some reason
```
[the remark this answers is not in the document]
alextabachnik: fixed

atolikin94 (approved):

kedeme (approved):

PR #125, opened 2023-10-05
Add support for include_meta_data to send the data only or send it with some meta data fields

ananalaghbar:


maor-walter:
[skip-ci] hpc-master migration to jenkins

PR #126, opened 2023-10-09
Issue: 3498200 Adding Retry mechanism for ufm slurm integration to

anasbadaha:
delete sharp reservation
Issue: 3547504: Change  default location for log_file_name in ufm_slurm.conf

### DIFF
--- a/plugins/SLURM-Integration/ufm_slurm_base.py
+++ b/plugins/SLURM-Integration/ufm_slurm_base.py
@@ -15,7 +15,7 @@
 # Author: Ibrahimbar
 # Author: Anas Badaha
 
-import sys
+import sys, time, http
 import argparse
 import logging
 from ufm_slurm_utils import UFM, GeneralUtils, Integration, Constants
@@ -53,6 +53,8 @@ class UfmSlurmBase():
             self.app_resources_limit = -1
         else:
             self.app_resources_limit = -1
+        self.num_of_retries = int(self.general_utils.get_conf_parameter_value(Constants.CONF_NUM_OF_RETRIES))
+        self.retry_interval = int(self.general_utils.get_conf_parameter_value(Constants.CONF_RETRY_INTERVAL))
         self.is_in_debug_mode = self.general_utils.is_debug_mode()
 
     def parse_args(self):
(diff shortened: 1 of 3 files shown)


plugins/SLURM-Integration/ufm_slurm_base.py
```diff
@@ -107,12 +109,35 @@ def create_sharp_allocation(self, job_id, job_nodes):
...
+                    logging.warning(f"Deleting sharp reservation failed, sharp reservation with app_id: "
+                                    f"{job_id} is not found!")
+                    break
+                if response.status_code == http.client.NO_CONTENT:
+                    logging.info(f"Deleting sharp reservation with app_id: {job_id} completed successfully.")
+                    break
+                else:
+                    logging.error(f"Deleting sharp reservation with app_id: {job_id} failed! "
+                                  f"status_code: {response.status_code}, response: {response.text}")
+                    if self.num_of_retries == 0:
```
mike-dubman, line 127: looks like bug, why checking self.num_of_retries but decrementing left_retries?
left_retries should be used all over the func

mike-dubman: plz add test for func

anasbadaha: No, it's not a bug. During the initialization, I set left_retries to self.num_of_retries because I want to maintain the original value of self.num_of_retries without decrementing it. This is necessary to handle the scenario where self.num_of_retries is set to 0, indicating that I need to retry the deletion indefinitely.

mike-dubman: so plz put comment describing that check if infinite or compare to constant == INFINITE
also, too many nested ```ifs```, plz simplify
why need ``else``` after ```break```?

anasbadaha: I will add comment as you requested and try to simplify the ifs as much as I can

plugins/SLURM-Integration/ufm_slurm_base.py
```diff
@@ -107,12 +109,44 @@ def create_sharp_allocation(self, job_id, job_nodes):
...
+                    # handle the scenario where num_of_retries in ufm_slurm.conf file is set to Non-zero value,
+                    # indicating that I need to retry the deletion equal to num_of_retries.
+                    elif left_retries > 0:
+                        logging.info(f"Retrying in {self.retry_interval} seconds... (Retries left: {left_retries})")
+                        left_retries -= 1
+                        time.sleep(self.retry_interval)
+                    # In case number of left_retries finished without succeeding to delete the sharp reservation, needs
+                    # to exit the while loop after adding an appropriate error message to log file.
+                    else:
+                        logging.error(f"No more retries. Giving up after {self.num_of_retries} retries to delete "
```
kedeme, line 145:
```suggestion
                        logging.error(f"No more retries. Failed to delete sharp reservation {job_id} after {self.num_of_retries} attempts."
```

kedeme (changes requested):
please add automated tests for the new retry mechanism as part of the new UFM SLURM Integration regression tests (on a separated PR)

PR #136, opened 2023-11-09
issue:3660200 implement logrotate for ndt plugin

alextabachnik:


atolikin94 (commented):
How does it actually work? Don't see that ndt_logrotate.conf is used somewhere.. Also please add some tests

egershonNvidia (approved):

plugins/UFM_NDT_Plugin/build/config/ndt_logrotate.conf
```diff
@@ -0,0 +1,9 @@
+/log/ndt.log {
+    compress
+    rotate 10
```
kedeme, line 3: rotate and size should be configurable in the NDT plugin configuration.

kedeme (changes requested):
please fix my comment.

plugins/UFM_NDT_Plugin/build/config/ndt_logrotate.conf
alextabachnik: ? how? it is external mechanism not related to plugin implementation.

kedeme (approved):
please set a meeting to discuss the long term solution for UFM plugins log rotation.

PR #143, opened 2023-12-12
issue:3702340 Add to NDT merger report information about nodes with

alextabachnik:
duplicated node description

### DIFF
--- a/plugins/UFM_NDT_Plugin/build/config/ndt.conf
+++ b/plugins/UFM_NDT_Plugin/build/config/ndt.conf
@@ -16,6 +16,8 @@ host_patterns: (^(SAT|DSM)\d+ ibp.*$)
 port_validation_sleep_interval=5
 # number of attempts to verify boundary port state - polling
 port_validation_number_of_attempts=5
+# Flag if to include information about nodes with duplicated node description in merger report
+check_duplicated_node_description=False
 # Address of server, to be used to send cable validation report request.
 # If address or host name not set - feature not enabled.
 # If address is "localhost" or "127.0.0.1" - request for cable validation report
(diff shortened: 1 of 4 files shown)


plugins/UFM_NDT_Plugin/build/config/ndt.conf
```diff
@@ -16,6 +16,8 @@ host_patterns: (^(SAT|DSM)\d+ ibp.*$)
 port_validation_sleep_interval=5
 # number of attempts to verify boundary port state - polling
 port_validation_number_of_attempts=5
+# Flag if to include information about nodes with duplicated node description in merger report
+check_duplicated_node_description=False
```
kedeme, line 20: don't we want to have this check by default?

alextabachnik: Eitan Zahavi said that we do not need this check at all (merger will work correct even with duplications BTW).
Alex Fok said that need to add this verification to report too. 
Any way this will not fail verification like duplicated GUIDs (it is critical and topoconfig in this case will not be created)

plugins/UFM_NDT_Plugin/ufm_sim_web_service/merger_resources.py
```diff
@@ -250,6 +253,14 @@ def run_ibdiagnet_ndt_compare(self, ndt_file_name):
                     report_content["error"] = "Report creation failed for %s: Filed to check duplicated GUIDs. File %s not exists" % (ndt_file_name, IBDIAGNET_LOG_FILE)
                     report_content["status"] = NDT_FILE_STATUS_VERIFICATION_FAILED
                     raise ValueError(report_content["error"])
+            if self.include_duplicated_nd:
+                # check for duplicated node descriptions
+                status, duplicated_nd = check_duplicated_node_descriptions()
+                if status and duplicated_nd:
+                    # in case of duplicated ND - include records in final report
+                    # (notification only - no report failure)
```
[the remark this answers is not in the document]
alextabachnik: It is not critical for merger - topoconfig will be created correct any way.
And we always will have many hosts with "Mellanox Technologies ...." description ... so verification will fail all the time - we will newer suceed.

drorlevy (approved):

PR #175, opened 2024-02-13
Issue: 3776089 Adding support for Kerberos Authentication in SLURM

anasbadaha:
integration

### DIFF
--- a/plugins/SLURM-Integration/ufm_slurm_epilog.py
+++ b/plugins/SLURM-Integration/ufm_slurm_epilog.py
@@ -42,9 +42,7 @@ class UfmSlurmEpilog(UfmSlurmBase):
                 self.session = self.ufm.getServerSession(auth_type=self.auth_type, token=token)
             elif self.auth_type == Constants.KERBEROS_AUTH:
                 principal_name = self.general_utils.get_conf_parameter_value(Constants.CONF_PRINCIPAL_NAME)
-                principal_pass = self.general_utils.get_conf_parameter_value(Constants.CONF_PRINCIPAL_PASSWORD)
-                self.session = self.ufm.getServerSession(auth_type=self.auth_type, principal_name=principal_name,
-                                                         principal_password=principal_pass if principal_pass else "")
+                self.session = self.ufm.getServerSession(auth_type=self.auth_type, principal_name=principal_name)
             else:
                 logging.error("auth_type in ufm_slurm.conf file must be one of the following (%s, %s, %s)" % (
                     Constants.BASIC_AUTH, Constants.TOKEN_AUTH, Constants.KERBEROS_AUTH))
--- a/plugins/SLURM-Integration/ufm_slurm_epilog.py
+++ b/plugins/SLURM-Integration/ufm_slurm_epilog.py
@@ -23,44 +23,11 @@ from ufm_slurm_base import UfmSlurmBase
 class UfmSlurmEpilog(UfmSlurmBase):
 
     def epilog_init(self):
+        UfmSlurmBase.init(self)
         try:
-            self.auth_type = self.general_utils.get_conf_parameter_value(Constants.AUTH_TYPE)
-            if self.auth_type == Constants.BASIC_AUTH:
-                user2 = self.general_utils.get_conf_parameter_value(Constants.CONF_UFM_USER)
-                password2 = self.general_utils.get_conf_parameter_value(Constants.CONF_UFM_PASSWORD)
-                if user2 is None or password2 is None:
-                    logging.error("For using %s you should set a valid %s %s in ufm_slurm.conf" % (
-                        Constants.BASIC_AUTH, Constants.CONF_UFM_USER, Constants.CONF_UFM_PASSWORD))
-                    sys.exit(self.should_fail)
-                self.session = self.ufm.getServerSession(auth_type=self.auth_type, username=user2, password=password2)
-            elif self.auth_type == Constants.TOKEN_AUTH:
-                token = self.general_utils.get_conf_parameter_value(Constants.CONF_TOKEN)
-                if token is None:
-                    logging.error("For using %s you should set a valid %s in ufm_slurm.conf" % (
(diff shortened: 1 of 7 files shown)


plugins/SLURM-Integration/ufm_slurm_prolog.py
```diff
@@ -45,9 +45,14 @@ def prolog_init(self):
                         Constants.TOKEN_AUTH, Constants.CONF_TOKEN))
                     sys.exit(self.should_fail)
                 self.session = self.ufm.getServerSession(auth_type=self.auth_type, token=token)
+            elif self.auth_type == Constants.KERBEROS_AUTH:
```
kedeme, line 48: you are duplicating the logic from ufm_slurm_epilog - please define a function at UfmSlurmBase and reuse it...

plugins/SLURM-Integration/ufm_slurm_prolog.py
```diff
@@ -45,9 +45,14 @@ def prolog_init(self):
...
+                principal_name = self.general_utils.get_conf_parameter_value(Constants.CONF_PRINCIPAL_NAME)
+                principal_pass = self.general_utils.get_conf_parameter_value(Constants.CONF_PRINCIPAL_PASSWORD)
+                self.session = self.ufm.getServerSession(auth_type=self.auth_type, principal_name=principal_name,
+                                                         principal_password=principal_pass if principal_pass else "")
             else:
-                logging.error("auth_type in ufm_slurm.conf file must be one of the following (%s, %s)" % (
-                    Constants.BASIC_AUTH, Constants.TOKEN_AUTH))
+                logging.error("auth_type in ufm_slurm.conf file must be one of the following (%s, %s, %s)" % (
```
kedeme, line 54: duplicate error message from ufm_slurm_epilog - please add some string which identify the error message source (epilog / prolog)

kedeme (changes requested):
see my changes and fix...

plugins/SLURM-Integration/ufm_slurm_epilog.py
```diff
@@ -40,9 +40,14 @@ def epilog_init(self):
                         Constants.TOKEN_AUTH, Constants.CONF_TOKEN))
                     sys.exit(self.should_fail)
                 self.session = self.ufm.getServerSession(auth_type=self.auth_type, token=token)
+            elif self.auth_type == Constants.KERBEROS_AUTH:
+                principal_name = self.general_utils.get_conf_parameter_value(Constants.CONF_PRINCIPAL_NAME)
+                principal_pass = self.general_utils.get_conf_parameter_value(Constants.CONF_PRINCIPAL_PASSWORD)
```
alextabachnik, line 45: password will be stored as text in config file?
I don't think so ... should be encrypted somehow

anasbadaha: one of the optionsto to obtain the TGT  is to use the keytab method, by this way the user will not need to provide the password.

plugins/SLURM-Integration/ufm_slurm_prolog.py
anasbadaha: there is nothing that we can do her, because the epilog and prolog are two separated scripts, and for both of them we need to get principal name and password in order to get the session.

plugins/SLURM-Integration/ufm_slurm_prolog.py
anasbadaha: it's clear in the logs file from which script the logs are coming.

plugins/SLURM-Integration/ufm_slurm_utils.py
```diff
@@ -187,28 +191,32 @@ class UFM:
...
-                {'Content-Type': 'application/json; charset=utf-8'})
         elif auth_type == Constants.TOKEN_AUTH:
-            session.verify = False
-            session.headers.update(
-                {'Content-Type': 'application/json; charset=utf-8',
-                 'Authorization': 'Basic %s' % token})
+            session.headers.update({'Authorization': 'Basic %s' % token})
+        elif auth_type == Constants.KERBEROS_AUTH:
+            __, krb_context = kerberos.authGSSClientInit("HTTP", principal_name)
+            kerberos.authGSSClientStep(krb_context, principal_password)
```
Abeer-Moghrabi, line 217: with user principal, did you try to authenticate without adding the password in the config file, considering that the ticket was already generated?

anasbadaha: I tried it without sending the password and saw that it's working without any issues! because we are using principal name only with kerberos authentication after generating the TGT, So the password is redundant I will remove it form the ufm_slurm.conf file.

plugins/SLURM-Integration/ufm_slurm_epilog.py
anasbadaha: I have removed the usage of principal_pass from the conf file because it's not needed.

Abeer-Moghrabi (approved):

plugins/SLURM-Integration/ufm_slurm_prolog.py
anasbadaha: fixes

plugins/SLURM-Integration/ufm_slurm_utils.py
```diff
@@ -101,7 +101,8 @@ def getSlurmConfFile(self):
             path = path.split('=')[1]
             return "%s/%s" % (os.path.dirname(path), Constants.UFM_SLURM_CONF_NAME)
         else:
-            return "%s/%s" % (Constants.SLURM_DEF_PATH, Constants.UFM_SLURM_CONF_NAME)
+            # return "%s/%s" % (Constants.SLURM_DEF_PATH, Constants.UFM_SLURM_CONF_NAME)
```
kedeme, line 104: please uncomment...

anasbadaha: fixed

kedeme (approved):

PR #176, opened 2024-02-18
Fixing 3783470: [UFM - TFS plugin] : the streaming will be failed in case the content-length header is missing

ananalaghbar:
In case the content-length header is missing from the response headers, then the expected content length is equal to the actual content length


nasraj (commented):
Looks good. Thanks

nasraj (approved):

PR #182, opened 2024-03-11
issue:3814411: Integrate the TFS with the clxci C Fluent streamer instead of the current Python streamer

ananalaghbar:
## What
This PR is integrating the [C fluent writer](https://gitlab-master.nvidia.com/telemetry/Collectx/-/blob/master/server/exporter/fluentbit_writer.py) that has been implemented by the collectx team

## Why ?
Using the C fluent writer instead of the current [Python fluent sender](https://github.com/Mellanox-lab/fluent-logger-python) improving the overall streaming pipeline by ~30%

By default, the TFS will use the new C writer to stream the data via the forward Fluent's plugin
it could be changed via the configurations to set it back & use the current python's sender


Abeer-Moghrabi (approved):

nasraj (approved):
looks good. thanks

TalLerner (approved):

PR #190, opened 2024-04-15
issue: 3865227: Fixing [UFM - TFS plugin] : Streaming into a Fluent server using the HTTP via IPv6 isn't working

ananalaghbar:
## What ?
The streaming in the TFS isn't working using the HTTP protocol

## Why ?
Fixing [3865227](https://redmine.mellanox.com/issues/3865227)

## How ?
By adding the missing IPv6 brackets to the HTTP request URL


ananalaghbar:
bot:retest

vg12345 (approved):

drorlevy (approved):
i think we can set the URL in an _init_ function or designated function so we won't build the string every iteration

PR #192, opened 2024-04-15
issue:3862873: Fixing Fluent streamer doesn’t work in case of UFM docker

ananalaghbar:
## What ?
The streaming in the TFS isn't working using the Fluent C Forward protocol in the case of UFM Docker

## Why ?
Fixing [3862873](https://redmine.mellanox.com/issues/3862873)

## How ?
Currently, In the case of docker, the path `/opt/ufm/telemetry/collectx/lib/`doesn't exist, and this directory used as shared volume with the TFS to share the clx .so libraries that are required for the streaming

To fix the issue, need to get rid of any UFM dependencies, the libraries should be part of the TFS image instead of loading them from the shared volume.


plugins/fluentd_telemetry_plugin/build/Dockerfile
```diff
@@ -12,6 +13,8 @@ COPY ${SRC_BASE_DIR}/conf/supervisord.conf /etc/supervisor/conf.d/
 COPY ${SRC_BASE_DIR}/scripts/init.sh ${SRC_BASE_DIR}/scripts/deinit.sh /
 COPY ${SRC_BASE_DIR}/conf/tfs_httpd_proxy.conf ${SRC_BASE_DIR}/conf/fluentd_telemetry_plugin.cfg ${SRC_BASE_DIR}/conf/fluentd.conf ${BASE_PATH}/
 
+COPY ${SRC_BASE_DIR}/lib/libfluent-bit.so ${SRC_BASE_DIR}/lib/libraw_msgpack_api.so  ${TELEMETRY_BASE_PATH}/collectx/lib/
```
drorlevy, line 16: if you copy executables you probably bind to a certain telemetry version. did you consult the Telemetry team?

ananalaghbar: I chose a specific version of the executables because it's fully tested with TFS and changing versions could require TFS adjustments due to dependency changes,

TalLerner (approved):

PR #196, opened 2024-04-26
Fix for "Issue:3868454:Send a flag to disable the dynamic telemetry for the PDR"

vg12345:
Reverts Mellanox/ufm_sdk_3.0#195


vg12345:
"is_registered_discovery" flag (regardless to its value) causes the failure of dynamic telemetry start with error 400.

PR #200, opened 2024-05-02
Adding batch isolate devices

boazhaim:
Adding an option to keep only one port as healthy for a given switch.
This can be done in bulk based on the input CSV file.

This script was tested in several ways:
1. With one switch to isolate.
2. With two batches, each with one switch to isolate.
3. With a single batch of 3 devices, ensure the ports isolation requested was sent to the SM.


kedeme (approved):

kedeme (commented):
please add info how was the script tested.

alexfok (approved):

PR #208, opened 2024-05-07
Issue:3874733:Fixes for blacklist functionality in PDR plugin

vg12345:


plugins/pdr_deterministic_plugin/ufm_sim_web_service/api/pdr_plugin_api.py
```diff
@@ -46,51 +47,60 @@ def get_excluded_ports(self):
...
         """
-        data_dict = request.form.to_dict()
-        if not data_dict:
-            return "No ports added to exclude list"
-      
+
+        try:
+            pairs = self.get_request_data()
+        except JSONDecodeError:
+            return "Incorrect input format\n", HTTPStatus.BAD_REQUEST
```
TalLerner, line 64: I think the error string message should be in a constant since it appears more then once.

TalLerner, line 64: Also the \n should be in a constant

vg12345: Done

vg12345: Done

TalLerner (approved):

PR #221, opened 2024-06-02
issue: 3932766: [TFS Plugin]: Fixing The meta fields are broken when streaming.stream_only_new_samples flag is false

ananalaghbar:
## What
The meta fields in the TFS are broken in the case of streaming.stream_only_new_samples is false

## Why ?
The root cause for this is appending the meta fields will occur before preparing the full port_record object, it's an indentation issue


vg12345 (approved):

ananalaghbar:
fixed in https://github.com/Mellanox/ufm_sdk_3.0/pull/261

PR #223, opened 2024-06-26
issue: 3943982: unifying the authentication service between the UFM host & UFM consumer plugin

ananalaghbar:
## What
This PR will mainly extend the UFM host's Apache configurations to forward the consumer plugin's requests to the consumer's server directly, instead of using the ModelMain UFM plugin proxy

## Why ?
To fix [UFM Server] Bug SW [#3943041](https://redmine.mellanox.com/issues/3943041): Consumer plugin - enabling any of authentication methods (e.g. client cert.) on the UFM provider host won't be applied on the UFM consumer plugin

## How ?
By adding a new Location to the Apache that does the proxy to the consumer plugin's server instead of the modelmain port
The new location is protected using the UFM host's authentication server.
with this change, both applications (main UFM host & Consumer UFM) are protected via one centralized authentication service 

**The new file [ufm_plugin_ufm_consumer_httpd.conf](https://github.com/Mellanox/ufm_sdk_3.0/pull/223/files#diff-4b80af4f1c8fe35fdb51634b81913c11712d717aca58b223363a10c8df0f310b) is added temporarily, the plan is to have this configurations shared and managed by the plugins manager for all plugins**

## Testing ?
Manual Testing, by adding the new consumer plugin image to the UFM, trying to access the UFM consumer GUI/ REST API using different authentication methods (e.g. basic/session + client cert., tokens)

**This fix will only work with the client cert. when the authentication service is enabled.**

### DIFF
--- a/plugins/ufm_consumer_plugin/scripts/init.sh
+++ b/plugins/ufm_consumer_plugin/scripts/init.sh
@@ -34,7 +34,11 @@ update_http_apache_port() {
 
 echo /opt/ufm/files/licenses:/opt/ufm/files/licenses > /config/${PLUGIN_NAME}_shared_volumes.conf
 
-cp $SRC_DIR_PATH/conf/${PLUGIN_NAME}_httpd_proxy.conf $SRC_DIR_PATH/conf/${PLUGIN_NAME}_plugin.conf $SRC_DIR_PATH/conf/${PLUGIN_NAME}_ui_conf.json $SRC_DIR_PATH/conf/ufm_plugin_${PLUGIN_NAME}_httpd.conf ${CONFIG_PATH}
+cp $SRC_DIR_PATH/conf/${PLUGIN_NAME}_httpd_proxy.conf \
+   $SRC_DIR_PATH/conf/${PLUGIN_NAME}_plugin.conf \
+   $SRC_DIR_PATH/conf/${PLUGIN_NAME}_ui_conf.json \
+   $SRC_DIR_PATH/conf/ufm_plugin_${PLUGIN_NAME}_httpd.conf \
+   ${CONFIG_PATH}
 
 update_http_apache_port
 
--- a/plugins/ufm_consumer_plugin/scripts/init.sh
+++ b/plugins/ufm_consumer_plugin/scripts/init.sh
@@ -24,11 +24,7 @@ CONFIG_PATH=/config
 update_http_apache_port() {
   # update the plugin http port in the apache configurations
   port=8997 #default port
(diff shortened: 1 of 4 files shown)


alextabachnik (approved):
Looks good for me, but I am less familiar with Apache configuration

plugins/ufm_consumer_plugin/scripts/init.sh
```diff
@@ -21,9 +21,22 @@ PLUGIN_NAME=ufm_consumer
...
+  else
+    echo "The file ${CONFIG_PATH}/${PLUGIN_NAME}_httpd_proxy.conf does not exist, setting the default port ${port}"
+  fi
+  sed -i "s/@@CONSUMER_REST_PORT@@/${port}/g" ${CONFIG_PATH}/ufm_plugin_${PLUGIN_NAME}_httpd.conf
+}
+
 echo /opt/ufm/files/licenses:/opt/ufm/files/licenses > /config/${PLUGIN_NAME}_shared_volumes.conf
 
-cp $SRC_DIR_PATH/conf/${PLUGIN_NAME}_httpd_proxy.conf $SRC_DIR_PATH/conf/${PLUGIN_NAME}_plugin.conf $SRC_DIR_PATH/conf/${PLUGIN_NAME}_ui_conf.json ${CONFIG_PATH}
+cp $SRC_DIR_PATH/conf/${PLUGIN_NAME}_httpd_proxy.conf $SRC_DIR_PATH/conf/${PLUGIN_NAME}_plugin.conf $SRC_DIR_PATH/conf/${PLUGIN_NAME}_ui_conf.json $SRC_DIR_PATH/conf/ufm_plugin_${PLUGIN_NAME}_httpd.conf ${CONFIG_PATH}
```
samerd, line 37: 1. split long line with `\`
2. you assumed above that `${PLUGIN_NAME}_httpd_proxy.conf` may not exist and now it's being copied without a check. either you should check as well or assume it exist and return an error above.

samerd (commented):
looks great.
please see comments below.

plugins/ufm_consumer_plugin/scripts/init.sh
ananalaghbar: 1. Done.
2. In `update_http_apache_port`, I use a default port if the file is missing. Is this sufficient, or should I raise an error instead?

ananalaghbar: regarding #2 I removed the unnecessary checks for the file existence

samerd (approved):
Well done!

TalLerner (approved):

PR #257, opened 2024-09-17
issue:3950870 PDR: send a reset REST API of a port after isolation

vg12345:
## What
Send a reset REST API of a port after isolation

## Testing ?
Manual simulated testing from sources

## Special triggers


vg12345:
Need to add (per port):
    - Max number of resets
    - Time interval after last reset that zeroes reset counter (reenables resets)

vg12345:
bot:retest

drorlevy (approved):
looks good! we need to add a test

plugins/pdr_deterministic_plugin/ufm_sim_web_service/isolation_mgr.py
```diff
@@ -97,6 +119,36 @@ def eval_isolation(self, port_name, cause):
...
+            return
+
+        # Update port resets history
+        reset_history = self.ports_resets.get(port_name)
+        if not reset_history:
+            reset_history = PortReset(port_name)
+            self.ports_resets[port_name] = reset_history
+
+        reset_history.reset_count += 1
+        reset_history.reset_time = datetime.now()
```
egershonNvidia, line 150: because you put the reset time here, which is excellent, you can use it to reset the restarts counter if enough time has passed, but it is only something to add in the future.

egershonNvidia (approved):

plugins/pdr_deterministic_plugin/ufm_sim_web_service/isolation_mgr.py
vg12345: The reset the restarts counter is performed during reset conditions check before the actual reset

PR #259, opened 2024-09-22
issue:4043499: Implement XDR Support for TFS plugin

ananalaghbar:
## What ?
This PR is for adding support for the XDR under TFS plugin

## Why ?
TFS currently supports collecting telemetry data periodically and forwarding it to fluentd for the normal ports.
This PR is adding support to configure the telemetry endpoint(s) to retrieve and forward the telemetry data including the XDR ports [Aggregation & Plane ports]

[4043499](https://redmine.mellanox.com/issues/4043499): Implement XDR Support for TFS plugin

## How ?
By adding the new following parameters to the telemetry endpoint's configurations:
1. `xdr_mode`: boolean, true/false
2. `xdr_ports_types`: list of supported port types that we need to collect/forward them fluentd [legacy, aggregated, plane]

both parameters are set per each endpoint. could be updated/retrieved via the existing `/conf` API

## Testing ?
Tested manually, this change should be tested e2e once we have XDR simulation for the telemetry HTTP endpoint


egershonNvidia (approved):

plugins/fluentd_telemetry_plugin/src/streamer.py
```diff
@@ -438,9 +505,20 @@ def _parse_telemetry_csv_metrics_to_json_with_delta(self, data):  # pylint: disa
         is_meta_fields_available = len(self.meta_fields[0]) or len(self.meta_fields[1])
         output = []
 
-        port_id_keys_indices = []
-        for port_id_key in self.port_id_keys:
-            port_id_keys_indices += [i for i, x in enumerate(keys) if x == port_id_key]
+        normal_port_id_keys_indices = []
```
TalLerner, line 508: Shouldn't you check the xdr_modr here?
You can have different paths and save if checks

plugins/fluentd_telemetry_plugin/src/streamer.py
```diff
@@ -449,6 +527,11 @@ def _parse_telemetry_csv_metrics_to_json_with_delta(self, data):  # pylint: disa
             # skip the first row since it contains the headers
             # skip the last row since its empty row
             values = row.split(UFMTelemetryConstants.CSV_ROW_ATTRS_SEPARATOR)
+            port_id_keys_indices = normal_port_id_keys_indices
```
TalLerner, line 530: Here as well.

ananalaghbar:
bot:retest

plugins/fluentd_telemetry_plugin/src/streamer.py
ananalaghbar: The XDR headers, such as `port_type` and `sys_image_guid`, only appear when `xdr_mode` is true.
So, checking these headers is enough from my point of view without also checking the xdr_mode

TalLerner (approved):

PR #261, opened 2024-09-29
issue:4061560: [TFS plugin] When the telemetry is unresponsive stream cached data

ananalaghbar:
## What

This PR introduces a new fallback option to stream cached data from the last successful iteration when the telemetry endpoint is unavailable. It also includes several improvements to data caching:
- Centralized Port Key Logic: Moves the logic for generating port keys based on type to shared functions.
- Enhanced Caching: Enables caching even when delta mode is disabled.
- Fallback Streaming: Streams cached data if available when the telemetry endpoint is unresponsive.

## Why ?
The goal is to enhance user experience during telemetry endpoint downtimes
it's part of [MKT. UFM Telemetry] Feature Request [#3929853](https://redmine.mellanox.com/issues/3929853): [Bytedance NDR][Feature Request]: Request UFM TFS Telemetry data(PortXmitDataExtended) to avoid to drop to 0 for Switch

## How ?
Data fetched from telemetry endpoints will be stored (cached) in a map per telemetry. If a telemetry endpoint becomes unavailable, and there was a previous successful iteration, the TFS will stream the cached data instead of not pushing anything.
This option is configurable via a new option in the `/conf` API called `enable_cached_stream_on_telemetry_fail`, which is enabled by default.

## Testing ?
The following scenarios were manually tested:
1. Streaming data with delta mode enabled/disabled to ensure existing functionality is operational.
2. Stopping the telemetry to verify that cached data is forwarded correctly, regardless of delta mode status.
3. Disabling the `enable_cached_stream_on_telemetry_fail` option to confirm that no data is streamed when telemetry is down.

### DIFF
--- a/plugins/fluentd_telemetry_plugin/src/streamer.py
+++ b/plugins/fluentd_telemetry_plugin/src/streamer.py
@@ -134,6 +134,7 @@ class UFMTelemetryStreamingConfigParser(ConfigParser):
     STREAMING_SECTION_C_FLUENT__STREAMER = "c_fluent_streamer"
     STREAMING_SECTION_BULK_STREAMING = "bulk_streaming"
     STREAMING_SECTION_STREAM_ONLY_NEW_SAMPLES = "stream_only_new_samples"
+    STREAMING_SECTION_ENABLE_CACHED_STREAM_ON_TELEMETRY_FAIL = "enable_cached_stream_on_telemetry_fail"
     STREAMING_SECTION_ENABLED = "enabled"
 
     META_FIELDS_SECTION = "meta-fields"
@@ -199,6 +200,12 @@ class UFMTelemetryStreamingConfigParser(ConfigParser):
         return self.safe_get_bool(self.args.bulk_streaming,
                                   self.STREAMING_SECTION,
                                   self.STREAMING_SECTION_STREAM_ONLY_NEW_SAMPLES,
+                                  True)
+
+    def get_enable_cached_stream_on_telemetry_fail(self):
+        return self.safe_get_bool(None,
+                                  self.STREAMING_SECTION,
+                                  self.STREAMING_SECTION_ENABLE_CACHED_STREAM_ON_TELEMETRY_FAIL,
                                   True)
 
     def get_enable_streaming_flag(self):
@@ -278,7 +285,7 @@ class UFMTelemetryStreaming(Singleton):
             'mvcr_sensor_name': 'mvcr_sensor_name', 'mtmp_sensor_name': 'mtmp_sensor_name',
             'switch_serial_number': 'switch_serial_number', 'switch_part_number': 'switch_part_number'
         }
-        self.last_streamed_data_sample_per_port = {}
+        self.last_streamed_data_sample_per_endpoint = {}
 
         self.streaming_metrics_mgr = MonitorStreamingMgr()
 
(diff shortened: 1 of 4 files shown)


plugins/fluentd_telemetry_plugin/src/streamer.py
```diff
@@ -702,44 +865,59 @@ def stream_data(self, telemetry_endpoint):  # pylint: disable=too-many-locals
...
-            logging.error("Failed to get the telemetry data metrics for %s", _url)
+                logging.info('Processing of endpoint %s Completed In: %.2f Seconds. '
+                             '(%d) Ports, (%d) Counters Were Handled',
+                             msg_tag, resp_process_time, data_len, num_of_counters)
+            elif self.stream_cached_data_on_telemetry_fail_is_enabled and \
+                    self.last_streamed_data_sample_per_endpoint.get(msg_tag):
+                # if the telemetry endpoint currently unavailable, try to get the cached data
+                data_to_stream = list(self.last_streamed_data_sample_per_endpoint[msg_tag].values())
+                data_len = len(data_to_stream)
+                if data_len > 0:
```
TalLerner, line 896: why the condition is data_len > 0 and no data_len == 0 for unavailable endpoint?

plugins/fluentd_telemetry_plugin/src/streamer.py
```diff
@@ -702,44 +865,59 @@ def stream_data(self, telemetry_endpoint):  # pylint: disable=too-many-locals
...
+            data_to_stream = []
+            new_data_timestamp = None
+            num_of_counters = data_len = 0
+            if telemetry_data:
                 ufm_telemetry_is_prometheus_format = self._check_data_prometheus_format(telemetry_data)
                 logging.info('Start Processing The Received Response From %s', msg_tag)
                 start_time = time.time()
-                data_to_stream, new_data_timestamp, num_of_counters = self._parse_telemetry_prometheus_metrics_to_json(telemetry_data) \
+                data_to_stream, new_data_timestamp, num_of_counters = \
+                    self._parse_telemetry_prometheus_metrics_to_json(telemetry_data, msg_tag) \
```
TalLerner, line 879: There is no Prometheus with XDR?

plugins/fluentd_telemetry_plugin/src/streamer.py
vg12345: data_len > 0 seems to be check for last valid cached data that should be streamed if there is no telemetry.

vg12345 (approved):

plugins/fluentd_telemetry_plugin/src/streamer.py
ananalaghbar: Currently, there are some limitations with using XDR in the Prometheus format, as detailed in [#4096650](https://redmine.mellanox.com/issues/4096650). 
Additionally, in the context of TFS, the Prometheus format is not commonly used. Most use cases prefer the CSV format due to its smaller response size compared to Prometheus.

plugins/fluentd_telemetry_plugin/src/streamer.py
ananalaghbar: As @vg12345 mentioned, this check is applied to verify if there's cached data for the specific endpoint or not

nasraj (approved):

PR #286, opened 2024-12-05
issue: 4197110: Add the collectx versions to the to the PDF

Miryam-Schwartz:
## What
_Add the collectx versions to the to the PDF._ 

## Why ?
_Display of extracted data._

## How ?
_save_collectx_versions method was added to the class set _funcs_for_analysis. This method extracts the versions and save them in _dataframes_for_pdf list._

## Testing ?
_Tested manually to ensure the output looks the same as before the change._

## Special triggers

### DIFF
--- a/plugins/ufm_log_analyzer_plugin/src/loganalyze/log_analyzers/console_log_analyzer.py
+++ b/plugins/ufm_log_analyzer_plugin/src/loganalyze/log_analyzers/console_log_analyzer.py
@@ -107,7 +107,7 @@ class ConsoleLogAnalyzer(BaseAnalyzer):
             "Time",
             "Amount of exceptions",
             "Exceptions count",
-    )
+        )
 
     def save_ufm_versions(self):
         self._txt_for_pdf.append(
--- a/plugins/ufm_log_analyzer_plugin/src/loganalyze/log_analyzers/console_log_analyzer.py
+++ b/plugins/ufm_log_analyzer_plugin/src/loganalyze/log_analyzers/console_log_analyzer.py
@@ -28,7 +28,10 @@ class ConsoleLogAnalyzer(BaseAnalyzer):
         self.fix_lines_with_no_timestamp(logs_csvs)
         super().__init__(logs_csvs, hours, dest_image_path, sort_timestamp)
         self._log_data_sorted.dropna(subset=["data"], inplace=True)
-        self._funcs_for_analysis = {self.print_exceptions_per_time_count}
+        self._funcs_for_analysis = {
+            self.print_exceptions_per_time_count,
+            self.save_ufm_versions,
+        }
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


plugins/ufm_log_analyzer_plugin/src/loganalyze/log_analyzers/ibdiagnet2_port_counters_analyzer.py
```diff
@@ -52,11 +53,16 @@ def __init__(
...
 
-    def get_collectx_versions(self):
+    def save_collectx_versions(self):
         unique_collectx_versions = self._log_data_sorted[
             self._log_data_sorted["type"] == "collectx_version"
         ]["data"].unique()
-        return unique_collectx_versions
+        self._dataframes_for_pdf.append(
+            (
+                f"{self.telemetry_type} telemetry collectx versions",
```
boazhaim, line 62: collectx versions found in {self.telemetry_type} telemetry log

boazhaim (approved):

plugins/ufm_log_analyzer_plugin/src/loganalyze/pdf_creator.py
```diff
@@ -60,7 +60,8 @@ def add_images(self):
     def add_text(self):
         self.set_font("Arial", "", 12)
         output = StringIO()
-        print(self._fabric_stats_list, file=output)
+        for txt in self._txt_list:
+            print(txt, file=output)
```
boazhaim, line 64: If you touched it :) (Which is great!)
Why do we need this logic? Cant we add these texts and combine them with the text we send in add_list_of_dicts_as_text ? 
(will also mean a small refactor, but it makes sense)
Not a must right now...

plugins/ufm_log_analyzer_plugin/src/loganalyze/log_analyzers/console_log_analyzer.py
```diff
@@ -106,6 +109,11 @@ def print_exceptions_per_time_count(self):
             "Exceptions count",
         )
 
+    def save_ufm_versions(self):
+        self._txt_for_pdf.append(
+            f"Used ufm version in console log {self.ufm_versions}{os.linesep}"
```
boazhaim, line 114: With the change you did, I'm not sure we need the os.linesep here. And if we do, it does not make sense to have it here, I would move it to the PDF class

plugins/ufm_log_analyzer_plugin/src/loganalyze/log_analyzers/ibdiagnet2_port_counters_analyzer.py
Miryam-Schwartz: fixed

PR #293, opened 2025-01-27
issue: #4258688: TFS plugin 1.1.0.0 have issues with fetching data from multiple telemetry endpoints.

egershonNvidia:
## What
Fix the P1 bug on the TFS plugin related to multi-telemetry configuration.

## Why ?
[P2](https://redmine.mellanox.com/issues/4258688)
Index out-of-range exception error for configuration setting of multi-telemetry.

## How ?
I have added a check for every item we use inside the multi-telemetry configuration, collecting all the items that are not equal to the host's length. If one item is not equal, raise an exception.
Capture an exception in the set conf API to check that it does not fail when updating.

## Testing ?
Manual testing.

## Special triggers

### DIFF
--- a/plugins/fluentd_telemetry_plugin/src/streamer.py
+++ b/plugins/fluentd_telemetry_plugin/src/streamer.py
@@ -97,16 +97,15 @@ class UFMTelemetryStreaming(Singleton):
         intervals = self.streaming_interval.split(splitter)
         msg_tags = self.fluentd_msg_tag.split(splitter)
 
-        error_message = ""
+        items_missing_length = []
         expected_amount = len(hosts)
         for name, array in [("port", ports), ("url", urls), ("interval" ,intervals), ("message_tag_name", msg_tags),\
                              ("xdr_mode", xdr_mode), ("xdr_ports_types", xdr_ports_types)]:
             if len(array)!= expected_amount:
-                error_message += f"setting under ufm-telemetry-endpoint {name} for multi telemetry,"\
-                  +"expected {len(hosts)} options got {len(array)}."
-
-        if len(error_message)>0:
-            raise InvalidConfRequest(error_message)
+                items_missing_length.append(name)
+        if len(items_missing_length)>0:
+            raise InvalidConfRequest(f"setting under ufm-telemetry for the following field {items_missing_length}"\
+                f"for multi telemetry. they need to have {expected_amount}")
 
         endpoints = []
         for i, value in enumerate(hosts):
--- a/plugins/fluentd_telemetry_plugin/src/streamer.py
+++ b/plugins/fluentd_telemetry_plugin/src/streamer.py
@@ -28,11 +28,17 @@ from monitor_streaming_mgr import MonitorStreamingMgr
 from telemetry_attributes_manager import TelemetryAttributesManager
 from telemetry_constants import UFMTelemetryConstants
 from telemetry_parser import TelemetryParser
-from api import InvalidConfRequest
 
 # pylint: disable=no-name-in-module,import-error
 from utils.utils import Utils
 from utils.singleton import Singleton
+
+class InvalidConfSetting(Exception):
+    """InvalidConfSetting Exception class for problem with the configuration file or updating"""
+
+    def __init__(self, message):
+        Exception.__init__(self, message)
+
 
 #pylint: disable=too-many-instance-attributes
 class UFMTelemetryStreaming(Singleton):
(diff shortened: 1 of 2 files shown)


plugins/fluentd_telemetry_plugin/src/streamer.py
```diff
@@ -95,6 +96,18 @@ def ufm_telemetry_endpoints(self):
         xdr_ports_types = self.ufm_telemetry_xdr_ports_types.split(splitter)
         intervals = self.streaming_interval.split(splitter)
         msg_tags = self.fluentd_msg_tag.split(splitter)
+
+        error_message = ""
+        expected_amount = len(hosts)
+        for name, array in [("port", ports), ("url", urls), ("interval" ,intervals), ("message_tag_name", msg_tags),\
+                             ("xdr_mode", xdr_mode), ("xdr_ports_types", xdr_ports_types)]:
+            if len(array)!= expected_amount:
```
TalLerner, line 104: How are you checking the xdr_ports_types?
It can't be compared to the expected amount.

egershonNvidia: It is used in line 117.
I didnt want the change the given logic, but if we dont add this check there it will fail for index-out-of-range from the for loop and line 96.

TalLerner: I didn't see the erparators.

plugins/fluentd_telemetry_plugin/src/streamer.py
```diff
@@ -95,6 +96,18 @@ def ufm_telemetry_endpoints(self):
         xdr_ports_types = self.ufm_telemetry_xdr_ports_types.split(splitter)
         intervals = self.streaming_interval.split(splitter)
         msg_tags = self.fluentd_msg_tag.split(splitter)
+
+        error_message = ""
+        expected_amount = len(hosts)
+        for name, array in [("port", ports), ("url", urls), ("interval" ,intervals), ("message_tag_name", msg_tags),\
```
ananalaghbar, line 102: I think you should validate only the required fields that are defined in the schema https://github.com/Mellanox/ufm_sdk_3.0/blob/main/plugins/fluentd_telemetry_plugin/src/schemas/set_conf.schema.json#L59

the optional fields such as the message_tag_name or xdr_mode should be included in the validation only if they have data

plugins/fluentd_telemetry_plugin/src/streamer.py
```diff
@@ -95,6 +96,18 @@ def ufm_telemetry_endpoints(self):
...
+                             ("xdr_mode", xdr_mode), ("xdr_ports_types", xdr_ports_types)]:
+            if len(array)!= expected_amount:
+                error_message += f"setting under ufm-telemetry-endpoint {name} for multi telemetry,"\
```
ananalaghbar, line 105: instead of repeating the same message for all invalid fields, I suggest you to change it into one message that contains all the invalid fields after validating them
```suggestion
                setting under ufm-telemetry-endpoint for the following fields {invalid_fields} for multi telemetry,....etc
```

egershonNvidia: agree, changed, but now I didnt write what they have, rather what the expected length they should be.

plugins/fluentd_telemetry_plugin/src/streamer.py
```diff
@@ -97,16 +97,15 @@ def ufm_telemetry_endpoints(self):
...
                              ("xdr_mode", xdr_mode), ("xdr_ports_types", xdr_ports_types)]:
             if len(array)!= expected_amount:
-                error_message += f"setting under ufm-telemetry-endpoint {name} for multi telemetry,"\
-                  +"expected {len(hosts)} options got {len(array)}."
-
-        if len(error_message)>0:
-            raise InvalidConfRequest(error_message)
+                items_missing_length.append(name)
+        if len(items_missing_length)>0:
+            raise InvalidConfRequest(f"setting under ufm-telemetry for the following field {items_missing_length}"\
```
ananalaghbar, line 107: 1. It is not correct to use `InvalidConfRequest` as this exception is related to the API and here the API isn't related.
2. Need to choose better phrases for the error message, please see the below suggestion:
`f"The following fields: {items_missing_length} in the ufm-telemetry-endpoint section are expected to contain comma-separated values with a length of {len(hosts)}. However, the provided values have a length of {len(array)}"`

egershonNvidia: 1. I agree, but I didnt want to create an exception for this, and I could not return [] because it would continue with the configuration as nothing is wrong, so I have to throw an exception and I saw this class, which is correct here as well, at least the name is, and it mostly occur in the conf_api.py which is what we using too.
2. sure, I will change to that.

ananalaghbar (approved):

PR #295, opened 2025-01-29
issue:4276633 Sys log also to redhat

boazhaim:
## What
Writing sys log messages also to redhat
## Why ?
To support all the OS platforms that UFM can run on

## Testing ?
Run on red had and seen the message in the syslog
## Special triggers


ananalaghbar (approved):

shmufeld (commented):
why not checking distribution then create the proper syslog?
also I think you should use macros for the syslog path

boazhaim:
Regarding the comment from shmuel - we talked.
While his idea is correct, we agreed that the current is better as it tries to use the 2 options we have, so we don't really need to know which operating system runs what.

PR #310, opened 2025-03-06
issue:4337421: [Plugins infra - ufm_sdk_3.0]: Replace singleton metaclass to the unified version from ufm_sdk_tools repo

ananalaghbar:
## What
Replace the current singleton utility in the `TFS`,`Bright` & `AdvancedHelloWorld` plugins  with the shared one that exists in `ufm_sdk_tools` 

## Why ?
Part of [UFM] Task [#4337413](https://redmine.mellanox.com/issues/4337413): [Plugins infra]: Unify and Centralize Singleton Implementation Across All Plugins


vg12345 (approved):

TalLerner (approved):

plugins/fluentd_telemetry_plugin/src/streamer.py
```diff
@@ -31,7 +31,7 @@
 
 # pylint: disable=no-name-in-module,import-error
 from utils.utils import Utils
-from utils.singleton import Singleton
+from ufm_sdk_tools.src.utils.singleton import SingletonMeta
```
drorlevy, line 34: don't you need to add it in the docker file as well since it is in a different location from your source code?

PR #325, opened 2025-04-29
issue:4413099 [BugFix] Added reboot event

atolikin94:
## What
Added reboot event in case switch was rebooted

## Testing ?
Manually rebooted the switch, got new reboot event, checked that other events are also being received

## Special triggers


plugins/gnmi_nvos_events_plugin/gnmi_events_server/gnmi_events_receiver.py
```diff
@@ -106,30 +107,46 @@ def subscribe(self, ip, user, credentials, guid, max_retries=1):
...
                             continue
                         self.traps_number += 1
                         payload = {"object_name": guid, "otype": "Switch"}
                         event_update = event_dict.get("update")
                         if event_update:
                             event_update = event_update.get("update")
-                            description = ""
-                            resource = ""
+                            description = resource = event_id = None
                             for event in event_update:
```
boazhaim, line 120: Won't it be better to iterate over reversed(my_list)?
This way, you will find the last one that matches first?

plugins/gnmi_nvos_events_plugin/gnmi_events_server/gnmi_events_receiver.py
```diff
@@ -106,30 +107,46 @@ def subscribe(self, ip, user, credentials, guid, max_retries=1):
...
+                                if show_last_event:
+                                    # iterate over stream of past events since they are unordered and save the last one
+                                    if description and resource and event_id and event_severity:
+                                        last_event_payload = payload.copy()
+                                        last_event_payload["description"] = f"{resource}: {description}" if resource else description
+                                        last_event_payload["event_id"] = event_severity
+                                        self.event_by_id[event_id] = last_event_payload
+                                        description = resource = event_id = None
+                            if show_last_event:
+                                sorted_events = dict(sorted(self.event_by_id.items()))
```
boazhaim, line 139: the sort is based on which key/info?

This is havey stuff just for finding one element.
Won't it be better to create it sorted from the start?
Or maybe to use SortedDict (if you can)

plugins/gnmi_nvos_events_plugin/gnmi_events_server/gnmi_events_receiver.py
atolikin94: the order is random, so we have to iterate over all of them

plugins/gnmi_nvos_events_plugin/gnmi_events_server/gnmi_events_receiver.py
atolikin94: looking at it now, I also think it's an overkill, even with SortedDict, since there can be lots of events on every switch. probably it's better to just trigger an event with custom text, something like "the switch was disconnected (e.g., because of reboot), connection restored"

shmufeld (approved):

boazhaim (dismissed):

PR #373, opened 2026-06-14
feature/ufm-state-mirror-sidecar

shmufeld:
## What
Adds `ufm-state-mirror/` to this repo: the StateMirror sidecar (Redis-backed UFM
storage) packaged as a first-class, **consumer-agnostic** component rather than
living inside the UFM repo.

- Engine package (`state_mirror`) — init-container restore (`state_mirror.restore`)
  and runtime sidecar (`state_mirror.mirror`), plus 126 unit tests.
- Self-contained image build: `build/Dockerfile` + `build/docker_build.sh`
  (image tag derived from the `VERSION` file).
- `VERSION` / `BUILD.md` release flow and component-scoped `ruff`/`pytest` config,
  mirroring the existing `ufm-plugin-helm-template` component pattern.
- GitHub Actions CI (`.github/workflows/state-mirror-ci.yml`): `ruff` lint/format,
  `pytest`, and a docker build (no push) on any change under `ufm-state-mirror/**`.
- The classifier is **decoupled from the image**: the engine reads `CLASSIFIER_PATH`
  at runtime, and the UFM file-set is kept here only as `examples/classifier-example.yaml`.

## Why ?
The sidecar is not a single-repo concern — it is intended to be shared by **both
the UFM and UFM HA charts** (separate repos). A component consumed by two repos
should not be buried inside one of them, otherwise the second consumer ends up
depending on the first and the two drift. Hosting it here as a standalone,
independently-versioned image lets both charts consume it by pinned tag. This
repo also already has a clean per-component image-build pattern (Harbor), so
building and maintaining the image here is simpler than wiring a new build into
the UFM appliance.

Relates to Redmine #5072180 (StateMirror sidecar) / #5023644 (remove UFM PVC,
push configuration/state to Redis). HLD: `HLD_redis_sidecar_storage_mode` (UFM repo).

## How ?
- **Standalone component, not a plugin.** It sits at the repo top level alongside
  `ufm-plugin-helm-template/` (its own `VERSION`/`BUILD.md`/CI), since it is an
  in-pod sidecar, not a REST-API plugin pod.
- **Engine vs. classifier split.** The image ships only the generic mirroring
  engine. Each consumer (UFM, UFM HA) mounts its own classifier ConfigMap at
  `CLASSIFIER_PATH` (default `/etc/state_mirror/state_mirror.yaml`). This keeps
  UFM-specific file-set knowledge — and its "reconcile against the real UFM tree"
  validation — in the consumer repo, while this repo stays vendor-neutral.
- **Image roles.** A single image runs both roles, selected by command: restore
  (init container) and mirror (native sidecar, the image default).
- The Docker build context is the component root so the existing `Dockerfile`
  `COPY` paths work unchanged.

## Testing ?
- **Unit tests:** 126 tests pass (`pytest`), covering classifier schema/validation,
  all handlers (`blob`, `atomic_blob`, `directory`, `sqlite`), wire format, watcher,
  health/metrics, redis error classification, and logging.
- **Lint/format:** `ruff check` and `ruff format --check` clean.
- **Image build:** verified the image builds from the committed snapshot via
  `build/docker_build.sh`.
- All of the above run automatically in the new GitHub Actions workflow on every PR.
- No E2E added in this PR — end-to-end wiring (the `ufm-enterprise`/UFM-HA charts
  pointing at the published image + shipping their classifier ConfigMap) is the
  follow-up consumer-side change.

## Special triggers


ufm-state-mirror/state_mirror/store.py
```diff
@@ -0,0 +1,292 @@
...
+
+        # Intentionally conservative: we count the base64 length of the body,
+        # which is ~33% larger than the raw bytes etcd actually stores (base64 is
+        # only the REST/JSON representation of binaryData). This fails closed a bit
+        # early rather than risk an apiserver rejection near the ~1 MiB ceiling.
+        total = len(b64) + sum(len(k) + len(v) for k, v in data.items())
+        if total > self._max:
+            log.error("%s encoded object %d bytes exceeds limit %d", key, total, self._max)
+            raise wire.WireError(
+                f"{key}: encoded object {total} bytes exceeds ConfigMap limit {self._max}",
```
samerd, line 222: **Caveat: the default `configmap` backend fails closed at ~1 MiB, but the classifier documents `gv.db` as up to ~50 MiB.**

`ConfigMapStore.put()` raises `WireError` once an encoded object exceeds `DEFAULT_MAX_OBJECT_BYTES` (~1 MiB). The default backend is `configmap` (`backends.py`), and the example classifier lists `gv.db` under it while noting it is "typically < 50 MiB" (and telemetry history can grow too). So on a real deployment, as soon as `gv.db` / `ufm_telemetry.db` crosses ~1 MiB, every mirror cycle for that key fails closed — the largest and most important state silently stops being protected.

Could this be a hard, up-front requirement instead of a comment? e.g. (a) document that `configmap` is only viable when *every* classified object stays under ~1 MiB, and (b) recommend/require the `redis` backend for any deployment that mirrors `gv.db`/telemetry. A startup validation that rejects `configmap` + a `sqlite gv.db` entry would surface this at deploy time rather than as runtime `WireError`s.

(FYI on the k0s side we hit the same ConfigMap ceiling and chose to *skip* oversized files with a warning rather than fail closed — different tradeoff, same underlying limit, so the doc guidance is the key thing.)

ufm-state-mirror/examples/classifier-example.yaml
```diff
@@ -0,0 +1,239 @@
...
+  - path: /opt/ufm/files/conf/host2port_guids_extended.json
+    handler: blob
+    redis_key: ufm:state:host2port_guids_extended
+    trigger: watchdog
+    baseline: skip
+    phase: 3
+
+  # Client certificate authentication store (JSON, despite the .db extension).
+  # Writer: gvvm/infra/client_authentication/ufm_client_authen.py _write_client_authen_data_to_db
+  - path: /opt/ufm/files/conf/webclient/ufm_client_authen.db
```
samerd, line 52: **Caveat: credential/user DBs are mirrored into plaintext ConfigMaps under the default backend.**

This entry, plus `gv.db` (holds users, line ~201) and `token_based_auth.db` (line ~222), are classified for the default `configmap` backend. ConfigMap values are base64 (`store.py`) but **not encrypted**, and there is no Secret/encryption path anywhere in `state_mirror/`. So auth material and API tokens end up in etcd effectively as plaintext, readable by anyone with `get configmap` in the namespace — typically a broader, more loosely-scoped permission than `get secret`.

Was Secret-backed storage for the credential-bearing entries considered? A per-entry `sensitive: true` that routes those keys to a Secret (leaning on etcd encryption-at-rest) would close the at-rest/RBAC gap. (We're adding this routing on the k0s side — happy to compare notes.)

ufm-state-mirror/state_mirror/handlers/directory.py
```diff
@@ -0,0 +1,115 @@
...
+        for relpath, full in self._iter_local_files():
+            try:
+                body = self._read_file(full)
+                key = self._key_for_rel(relpath)
+                if self._push_if_changed(key, body):
+                    log.info("mirror: shipped %s -> %s", full, key)
+                    sent_any = True
+            except Exception:
+                log.exception("mirror: failed to ship child %s; continuing", relpath)
+        return sent_any
```
samerd, line 89: **Caveat: per-child write failures are swallowed with no failure signal.**

`mirror()` catches `Exception` per child, logs `"…continuing"`, and returns `sent_any` — i.e. "did *anything* ship," not "did *everything* ship." If a child fails (apiserver error, or a child that trips the ~1 MiB limit and raises `WireError`), the cycle still reports success, the child stays stale/absent in the backend, and a later restore is silently incomplete. There is no per-child failure counter to alert on.

Suggestion: track a failure count and (a) expose it as a metric (e.g. `*_mirror_child_errors_total`) and/or (b) factor it into handler health, so a persistently failing child becomes visible instead of silent.

samerd (commented):
Thanks for this — overall a solid, well-documented design; the classifier-as-source-of-truth, WAL-aware sqlite snapshots, and fail-closed restore are all the right calls. I left three inline comments (ConfigMap size limit vs `gv.db`, plaintext credential DBs, and swallowed directory child failures). A few smaller, non-blocking notes for the record:

1. **Whole-DB snapshot write amplification.** `sqlite` mirrors ship the entire DB on any change-counter bump; for a busy multi-MiB `gv.db` on a short `poll_interval_ms`, that is repeated full-DB writes into the backend. A note on sizing `poll_interval_ms` vs DB size (especially on Redis) would help.

2. **Directory dirty → O(N) re-scan.** A single child change drives a full `_iter_local_files()` walk + per-child change check. Fine for small conf trees; calling out that assumption would help.

3. **A delete that happens while the mirror is down comes back.** The sidecar only removes a file from the backend if it *sees* the delete. If the sidecar/pod is down when a file is deleted, that delete is never recorded, and restore copies the file back on restart — silently undoing the deletion. This is the intended "backend wins" policy; just worth stating that outage-time deletes are not honored until re-observed.

4. **Anything missing from the classifier is lost on every pod restart.** With `emptyDir` runtime storage, the backend copy is the only copy. A UFM file that is not in the classifier is not mirrored and disappears on the next restart with no warning, so classifier completeness is critical. A report of files found under the watched roots with no classifier entry would help catch gaps.

5. **One bad file blocks the entire restore.** Restore fails closed: a single corrupt/oversized key aborts the whole restore and the pod will not come up, even if every other file is fine. Safe default — but is a "restore everything else and flag the bad key" mode in scope, so one bad entry cannot hold the instance hostage?

6. **Single-writer `replace` assumption.** Atomic `tmp+rename` assumes one writer per path; if two writers ever target the same file, last-writer-wins applies. Probably intentional — just confirming.

7. **No E2E yet.** Is an integration test planned that exercises mirror→kill→restore against a real (or kind) apiserver/Redis, including the sqlite and directory handlers?

mprivorotsky:
what about ci which is running on Ubuntu 24 but container itself is python slim which is Debian 13.5 Trixie?

PR #377, opened 2026-06-17
state-mirror (3/4): handlers + routing

shmufeld:
Third slice of the StateMirror sidecar (stacked on PR 2/4).

Adds the per-file-type mirror/restore logic and the wiring that binds it to a
backend:
- handlers: BaseHandler plus blob, atomic_blob, directory (per-child keys,
  realpath-containment guard on restore) and sqlite (consistent online-backup
  snapshot, WAL-aware change signature, fail-closed integrity check on
  restore). _write_file preserves the destination's uid/gid/mode across the
  tmp+rename so a root-run restore does not strip ownership.
- backends: install-wide backend selection (STATE_MIRROR_BACKEND) building a
  single Store at startup and binding each classifier entry's handler to it.

The mirror/restore entrypoints and health/metrics server follow in PR 4/4.
Tests: handlers, sqlite, backend routing; ruff clean.

## What
_Describe what this PR is doing._ 

## Why ?
_Justification for the PR. If there is existing issue/bug, please reference. For
Bug fixes why and what can be merged in a single item._

## How ?
_It is optional, but for complex PRs please provide information about the design,
architecture, approach, etc._

## Testing ?
_How was this feature tested? Did you add unit tests? Did you add E2E tests?_

## Special triggers

### DIFF
--- a/ufm-state-mirror/state_mirror/handlers/base.py
+++ b/ufm-state-mirror/state_mirror/handlers/base.py
@@ -104,7 +104,16 @@ class BaseHandler:
         return sent
 
     def on_delete(self) -> None:
-        """Propagate a local delete to the store (HLD 5.3.9)."""
+        """Propagate a local delete to the store (HLD 5.3.9).
+
+        Deletes are NOT durable across restarts for entries with
+        ``baseline=image`` or ``baseline=empty``: dropping the key means the next
+        startup finds no stored object, :meth:`restore` returns ``False``, and
+        ``bootstrap`` re-seeds the file from its baseline -- so a deliberately
+        deleted file reappears. This is consistent with the "backend wins on
+        ambiguity" policy; making deletes survive a restore would require writing
+        a tombstone key (future work).
+        """
         log.info("on_delete: dropping %s from store", self.entry.redis_key)
         self.store.delete(self.entry.redis_key)
 
@@ -221,9 +230,7 @@ class BaseHandler:
         try:
             os.chown(path, prev.st_uid, prev.st_gid)
         except OSError as exc:
-            log.warning(
-                "chown of %s to %d:%d failed: %s", path, prev.st_uid, prev.st_gid, exc
-            )
+            log.warning("chown of %s to %d:%d failed: %s", path, prev.st_uid, prev.st_gid, exc)
 
     @staticmethod
     def _restore_one(store: Store, key: str, dest_path: str) -> Optional[wire.Meta]:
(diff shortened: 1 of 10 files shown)


ufm-state-mirror/state_mirror/handlers/sqlite.py
```diff
@@ -0,0 +1,201 @@
...
+
+        Fails closed (raises) on an integrity-check failure so the init
+        container never lets UFM start on a corrupt DB.
+        """
+        result = self.store.get(self.entry.redis_key)
+        if result is None:
+            log.debug("restore: no stored object for %s", self.entry.path)
+            return False
+        body, _meta = result
+        self._write_file(self.entry.path, body)
```
vg12345, line 199: We overwrite the DB before `integrity_check()`. Could we check a temp copy first and `os.replace()` only if valid? A bad restore would then leave the old DB intact.

ufm-state-mirror/state_mirror/handlers/directory.py
```diff
@@ -0,0 +1,115 @@
...
+    def mirror(self) -> bool:
+        sent_any = False
+        for relpath, full in self._iter_local_files():
+            try:
+                body = self._read_file(full)
+                key = self._key_for_rel(relpath)
+                if self._push_if_changed(key, body):
+                    log.info("mirror: shipped %s -> %s", full, key)
+                    sent_any = True
+            except Exception:
```
vg12345: This also swallows backend write failures, but the loop still calls `record_store_ok()` — so an outage looks healthy. Could we let `WireError` bubble up and only skip a child on a local `OSError`?

ufm-state-mirror/state_mirror/handlers/base.py
```diff
@@ -0,0 +1,235 @@
...
+        if sent:
+            log.info(
+                "mirror: shipped %s (%d bytes) -> %s",
+                self.entry.path,
+                len(body),
+                self.entry.redis_key,
+            )
+        return sent
+
+    def on_delete(self) -> None:
```
vg12345, line 106: Delete just drops the key, so after restart a deleted file with a baked default can come back. Is backend-wins intended here, or should deletes survive restore (tombstone)?

alextabachnik (dismissed):

ufm-state-mirror/state_mirror/handlers/sqlite.py
shmufeld: Addressed. restore() verifies on a throwaway temp before touching the live DB:

ufm-state-mirror/state_mirror/handlers/directory.py
shmufeld: Addressed. mirror() now catches only OSError

ufm-state-mirror/state_mirror/handlers/base.py
shmufeld: Addressed (intentional, now documented).

alextabachnik (approved):

vg12345 (approved):

PR #378, opened 2026-06-17
state-mirror (4/4): runtime + health

shmufeld:
Final slice of the StateMirror sidecar (stacked on PR 3/4); completes the
component so the image is functional.

- mirror: the long-running sidecar loop -- watch (debounced) + periodic full
  scan, observed-delete reconcile (bounded, O(1) drop-oldest), unobserved
  drift kept-and-counted (backend wins on ambiguity), liveness heartbeat inside
  the scan/drain loops so a slow backend can't trip the probe and lose the
  emptyDir.
- restore: the one-shot init-container entrypoint (fail-closed).
- watcher: watchdog-based filesystem event handler + path resolution.
- health: readiness/liveness HTTP server and the Prometheus metric surface
  (backend-neutral names matching the chart's cross-repo contract).

After this PR the tree equals the original #373 branch.
Tests: full suite (mirror, watcher, health) green; ruff clean.

## What
_Describe what this PR is doing._ 

## Why ?
_Justification for the PR. If there is existing issue/bug, please reference. For
Bug fixes why and what can be merged in a single item._

## How ?
_It is optional, but for complex PRs please provide information about the design,
architecture, approach, etc._

## Testing ?
_How was this feature tested? Did you add unit tests? Did you add E2E tests?_

## Special triggers

### DIFF
--- a/ufm-state-mirror/state_mirror/watcher.py
+++ b/ufm-state-mirror/state_mirror/watcher.py
@@ -100,7 +100,23 @@ class MirrorEventHandler:
         if event_type == EVENT_CLOSED:
             self._dirty(event.src_path)
         elif event_type == EVENT_MOVED:
-            self._dirty(getattr(event, "dest_path", None) or event.src_path)
+            dest = getattr(event, "dest_path", None) or event.src_path
+            self._dirty(dest)
+            # The move is observed, so honor it: if the OLD path was a tracked
+            # entry (a renamed single file, or a child inside a watched dir), its
+            # backend key must be dropped or a later restore revives the file at
+            # the old location. Safe against atomic write-temp-then-rename -- the
+            # drain only deletes when the source path is actually gone and
+            # store.delete is idempotent, so a move-then-recreate is a no-op.
+            if event.src_path != dest:
+                src_entry = self._resolver.resolve(event.src_path)
+                if src_entry is not None:
+                    log.debug(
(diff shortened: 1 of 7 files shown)


ufm-state-mirror/state_mirror/watcher.py
```diff
@@ -0,0 +1,136 @@
...
+        self._mark_dirty = mark_dirty
+        self._mark_delete = mark_delete
+
+    def dispatch(self, event) -> None:
+        if getattr(event, "is_directory", False):
+            return
+        event_type = getattr(event, "event_type", None)
+        if event_type == EVENT_CLOSED:
+            self._dirty(event.src_path)
+        elif event_type == EVENT_MOVED:
```
vg12345, line 102: A move only marks the destination dirty, so the old key stays and can return on restore. Should we also queue a delete for `src_path` when it is a tracked entry?

alextabachnik (dismissed):

ufm-state-mirror/state_mirror/watcher.py
shmufeld: Done in d537bfa
dispatch() now marks the destination dirty and, when the old src_path resolves to a tracked entry (a renamed single file, or a child inside a watched dir)

vg12345 (approved):

alextabachnik (approved):

PR #381, opened 2026-07-06
issue:5134987 [BugFix] [1/3] Enforce state mirror startup correctness

shmufeld:
## What

Make StateMirror startup fail closed and make backend-health reporting reflect real mirror activity.

- Gate readiness on a successful initial reconcile and healthy required watchdog emitters.
- Keep readiness latched after startup while reporting runtime degradation.
- Add an explicit unsupported poll-only escape hatch.
- Distinguish writes, backend reads, and local no-ops.
- Preserve partial failures: backend failures win aggregate reachability, while
  local failures fail reconciliation without a false backend-outage signal.
- Treat ambiguous local inspection as a fail-closed reconcile error, including
  directory traversal, SQLite fingerprints, and pending-delete checks.
- Retain SQLite backend snapshots after local deletion and count once per onset.
- Correct dropped-delete metric documentation.

## Why ?

Redmine [#5134987](https://redmine.mellanox.com/issues/5134987), under story [#5079041](https://redmine.mellanox.com/issues/5079041).

The sidecar could report ready before mirroring and watchers were usable, and no-op or partial-result accounting could mask outages or incorrectly advance write-lag metrics.

## How ?

Handlers return an explicit `MirrorResult` containing write/read counts and failures. Drift scans report their actual backend activity instead of inferring it from racy path state. Startup retries until the initial reconcile, backend proof, and strict watcher-health gate all pass. Watcher health covers the dispatcher and every scheduled emitter.

This is **1/3** in a stacked series:

1. This PR — startup correctness and health accounting.
2. [PR #382](https://github.com/Mellanox/ufm_sdk_3.0/pull/382) — directory and classifier path safety.
3. [PR #383](https://github.com/Mellanox/ufm_sdk_3.0/pull/383) — SQLite replacement detection and logging parity.

## Testing ?

- Python 3.12: `ruff check .` — passed.
- Python 3.12: `ruff format --check .` — passed.
- Python 3.12: `python -m pytest -q` — **217 passed**.
- NVIDIA copyright validation — passed.
- Fresh-context review — no actionable correctness regressions.
- GitHub Actions Docker build — passed.

### DIFF
--- a/ufm-state-mirror/state_mirror/handlers/base.py
+++ b/ufm-state-mirror/state_mirror/handlers/base.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 
 import logging
 import os
+import sqlite3
 import stat
 from dataclasses import dataclass
 from enum import Enum
@@ -34,6 +35,20 @@ from state_mirror.classifier import Baseline, Entry
 from state_mirror.store import Store
 
 log = logging.getLogger(__name__)
+
+
+def path_exists(path: str) -> bool:
+    """Check presence without hiding local inspection failures.
+
+    ``os.path.exists`` converts every ``OSError`` to ``False``. That is too
+    permissive for reconciliation: only true absence is a local no-op;
+    permission and I/O failures must propagate and keep startup fail-closed.
+    """
+    try:
+        os.stat(path)
+        return True
+    except FileNotFoundError:
+        return False
 
 
 class MirrorOutcome(str, Enum):
@@ -51,7 +66,9 @@ class MirrorResult:
     Counts make directory aggregation unambiguous: writes dominate successful
     reads, which dominate a purely local no-op. Failures are deliberately kept
     separate so a partially successful directory pass can advance the last
-    write timestamp while still leaving final backend reachability false.
+    write timestamp while still failing the reconcile. Only failures caused by
+    backend activity affect backend reachability; local filesystem and SQLite
+    failures remain fail-closed for startup without misreporting the backend.
     """
 
     writes: int = 0
@@ -73,6 +90,25 @@ class MirrorResult:
     @property
     def succeeded(self) -> bool:
         return not self.failures
+
+    @staticmethod
+    def failure_affects_backend(exc: BaseException) -> bool:
+        """Return whether ``exc`` invalidates backend health."""
+        return (
+            not isinstance(exc, (OSError, sqlite3.Error))
+            and getattr(exc, "reason", None) != "local_io"
+        )
+
+    @property
+    def backend_failures(self) -> tuple[BaseException, ...]:
+        """Failures that mean the storage operation did not complete.
+
+        Local filesystem and SQLite errors prevent a complete reconcile, but
+        they say nothing about Redis or the Kubernetes API. Unknown failures
+        remain backend-affecting conservatively so an unclassified store error
+        cannot be reported as healthy.
+        """
+        return tuple(exc for exc in self.failures if self.failure_affects_backend(exc))
 
     def plus(self, other: "MirrorResult") -> "MirrorResult":
         """Combine sibling operations while retaining every failure."""
@@ -139,7 +175,7 @@ class BaseHandler:
         unchanged file is a cheap backend read; a missing local file is a true
         local no-op and does not imply anything about backend reachability.
         """
-        if not os.path.exists(self.entry.path):
+        if not path_exists(self.entry.path):
             log.debug("mirror: %s does not exist yet, skipping", self.entry.path)
             return MirrorResult()
         body = self._read_file(self.entry.path)
@@ -181,7 +217,7 @@ class BaseHandler:
         ambiguity (the next restore re-materializes the file). An empty list means
         no drift.
         """
-        if os.path.exists(self.entry.path):
+        if path_exists(self.entry.path):
             return []
         if self.store.get_meta(self.entry.redis_key) is None:
             return []
--- a/ufm-state-mirror/state_mirror/handlers/base.py
+++ b/ufm-state-mirror/state_mirror/handlers/base.py
@@ -217,11 +217,15 @@ class BaseHandler:
         ambiguity (the next restore re-materializes the file). An empty list means
         no drift.
         """
+        return self.drift_scan()[0]
+
+    def drift_scan(self) -> tuple[list[str], int]:
+        """Return drift keys and the number of backend reads performed."""
         if path_exists(self.entry.path):
-            return []
+            return [], 0
         if self.store.get_meta(self.entry.redis_key) is None:
-            return []
-        return [self.entry.redis_key]
+            return [], 1
+        return [self.entry.redis_key], 1
 
     def _push_if_changed(self, key: str, body: bytes) -> bool:
         meta = self.store.get_meta(key)
--- a/ufm-state-mirror/state_mirror/handlers/base.py
+++ b/ufm-state-mirror/state_mirror/handlers/base.py
@@ -32,6 +32,7 @@ from typing import Optional
 
 from state_mirror import wire
 from state_mirror.classifier import Baseline, Entry
+from state_mirror.redis_errors import LOCAL_IO_REASON
 from state_mirror.store import Store
 
 log = logging.getLogger(__name__)
@@ -96,7 +97,7 @@ class MirrorResult:
         """Return whether ``exc`` invalidates backend health."""
         return (
             not isinstance(exc, (OSError, sqlite3.Error))
-            and getattr(exc, "reason", None) != "local_io"
+            and getattr(exc, "reason", None) != LOCAL_IO_REASON
         )
 
     @property
--- a/ufm-state-mirror/state_mirror/handlers/base.py
+++ b/ufm-state-mirror/state_mirror/handlers/base.py
@@ -25,6 +25,8 @@ from __future__ import annotations
 import logging
 import os
 import stat
+from dataclasses import dataclass
+from enum import Enum
 from typing import Optional
 
 from state_mirror import wire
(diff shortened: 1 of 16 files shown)


alextabachnik (dismissed):

vg12345 (dismissed):

ufm-state-mirror/state_mirror/mirror.py
```diff
@@ -83,9 +84,10 @@ def __init__(
         # drift, never a delete (HLD 5.3.7/5.3.9, "the backend wins on ambiguity").
         self._pending_deletes: OrderedDict[str, tuple[str, str]] = OrderedDict()
         self._max_queue = max(1, max_queue)
-        # Keys already counted as unobserved drift (file gone but backend has it),
-        # so persistent drift bumps state_mirror_unexpected_delete_total once per
-        # onset rather than every scan (HLD 5.3.7).
+        # Keys already counted as missing-local drift. Usually the backend still
+        # has the object; an observed SQLite deletion is retained here even if
+        # its backend snapshot is absent, because only local reappearance ends
+        # that operator-error onset (HLD 5.3.7/5.3.9).
```
ananalaghbar, line 94: do we need the HLDs' references here?

ufm-state-mirror/state_mirror/handlers/base.py
```diff
@@ -34,6 +36,53 @@
...
+            return MirrorOutcome.WROTE
+        if self.reads:
+            return MirrorOutcome.READ_OK
+        return MirrorOutcome.LOCAL_NOOP
+
+    @property
+    def backend_ops(self) -> int:
+        return self.writes + self.reads
+
+    @property
```
ananalaghbar, line 91: A local filesystem error (e.g. a `conf/plugins/**` child that the non-root, read-only sidecar can't read, or a non-regular file) is raised as an `OSError` and ends up in failures. `classify_redis_error` already labels these local_io ("not a Redis fault"), but succeeded/reachability still treat them as a backend outage. That has two bad effects: (a) `state_mirror_backend_reachable` flips to 0 and `StateMirrorBackendUnreachable` fires even though Redis/etcd is healthy, and (b) at startup `full_scan().succeeded` is `False`, so startup_once() keeps retrying and the pod never becomes Ready — a condition the sidecar can't fix by itself.

Suggest adding a helper that separates true backend failures from local ones, then using it for the reachability/readiness gates (keep counting local_io in backend_errors_total for visibility).


```python
    @property
    def succeeded(self) -> bool:
        return not self.failures
 
   @property
    def backend_failures(self) -> tuple[BaseException, ...]:
        # OSError == local filesystem problem (classified "local_io"), NOT a
        # backend outage, so it must not flip reachability or block readiness.
        return tuple(e for e in self.failures if not isinstance(e, OSError))
```

ufm-state-mirror/state_mirror/mirror.py
```diff
@@ -211,56 +244,99 @@ def _apply_delete(self, entry_path: str, fs_path: str) -> None:
...
-                log.exception("mirror failed for %s [%s]", handler.entry.path, reason)
-                self.state.record_store_down(reason)
+            result = self._mirror_handler(handler)
+            self._record_result(result)
+            aggregate = aggregate.plus(result)
         flushed = self.flush_pending_deletes()
-        self._scan_unexpected_deletes()
+        aggregate = aggregate.plus(flushed)
+        aggregate = aggregate.plus(self._scan_unexpected_deletes())
+        if aggregate.failures:
```
ananalaghbar: Gate reachability on real backend failures only, so a local local_io error doesn't mark the backend unreachable (and doesn't block startup, since full_scan()'s result feeds startup_once).

```suggestion
        if aggregate.backend_failures:
            # Only a real backend failure wins final reachability for this
            # reconcile; local_io errors are recorded but the backend is fine.
            self.state.mark_backend_unreachable()
```

ufm-state-mirror/state_mirror/mirror.py
```diff
@@ -154,36 +181,41 @@ def drain_once(self, now: float, rate_limited: bool = True) -> int:
...
+            else:
+                self._last_ship[path] = now
         if requeue:
             with self._lock:
                 self._dirty |= requeue
-        ops += self.flush_pending_deletes()
+        deleted = self.flush_pending_deletes()
+        aggregate = aggregate.plus(deleted)
+        ops += deleted.writes
+        if aggregate.failures:
```
ananalaghbar: Same as `full_scan`: use `backend_failures` here so a local read error on one dirty file doesn't flip backend_reachable.


```suggestion
        if aggregate.backend_failures:
```

ufm-state-mirror/state_mirror/mirror.py
```diff
@@ -288,36 +371,123 @@ def poll_sqlite(self, now: float | None = None) -> int:
...
+            self._record_result(result)
+            aggregate = aggregate.plus(result)
+            sent += result.writes
+            if handler.last_snapshot_seconds is not None:
+                self.state.set_snapshot_duration(
+                    os.path.basename(path), handler.last_snapshot_seconds
+                )
         if sent:
             self.state.add_mirror_ops(sent)
+        if aggregate.failures:
```
ananalaghbar: Same fix here for consistency.


```suggestion
        if aggregate.backend_failures:
```

ufm-state-mirror/state_mirror/mirror.py
```diff
@@ -288,36 +371,123 @@ def poll_sqlite(self, now: float | None = None) -> int:
...
+        watchers_ok = self._observer_healthy(self._observer)
+        if not watchers_ok and self._observer is not None:
+            self._cleanup_observer(self._observer)
+            self._observer = None
+            self.state.mark_watching(False, all_required_armed=False)
+        if watchers_ok:
+            self.state.mark_watching(True, all_required_armed=True)
+        if not watchers_ok:
+            watchers_ok = self.start_watching()
+        result = self.full_scan()
```
ananalaghbar, line 510: The readiness gate uses `result.succeeded`, which includes `local_io`. If a single plugin child is unreadable, `startup_once()` returns `False` and run_forever() loops forever -> pod never Ready. Please base the gate on backend failures instead (a local error should still let the pod start; the file just isn't mirrored and is visible via backend_errors_total{reason="local_io"} + logs).

Concretely: compute `reconcile_ok = not result.backend_failures` after the probe block and use it at the two places that currently read `result.succeeded`:

```python
    # after the post-scan watcher recheck:
    reconcile_ok = not result.backend_failures
    ...
    self.state.mark_initial_reconcile(reconcile_ok)
    return reconcile_ok and watcher_gate
```

(The zero-op probe branch at line 462 can also use reconcile_ok so a local-only error still allows the backend probe to run.)

ufm-state-mirror/state_mirror/mirror.py
```diff
@@ -288,36 +371,123 @@ def poll_sqlite(self, now: float | None = None) -> int:
...
+            self._cleanup_observer(self._observer)
+            self._observer = None
+        self.state.mark_watching(watchers_ok, all_required_armed=watchers_ok)
+        if not watchers_ok and self.state.allow_poll_only:
+            log.warning("UNSUPPORTED poll-only mode enabled; readiness may pass without watchdog")
+        watcher_gate = watchers_ok or self.state.allow_poll_only
+        self.state.mark_initial_reconcile(result.succeeded)
+        return result.succeeded and watcher_gate
+
+    def run_forever(
```
ananalaghbar, line 534: There's no `SIGTERM` handler, so the loop is killed mid-flight and any change queued in the last `poll_interval (0.5s)` can be lost on pod deletion. Since native sidecars terminate after the main container, a small handler that runs a final `drain_once()` + `full_scan()` on `SIGTERM` would tighten the "latest local change wins" guarantee. Fine as a follow-up if out of scope here.

ananalaghbar (commented):
Main comment - local filesystem errors shouldn't count as a backend outage

This is one issue with several touch-points. I posted the root-cause comment once, then linked the small ones to it.

ufm-state-mirror/state_mirror/mirror.py
shmufeld: Implemented. full_scan() now uses aggregate.backend_failures for final backend reachability. Local failures remain in the aggregate so startup reconciliation still fails closed, but they do not fire the backend-unreachable signal.

ufm-state-mirror/state_mirror/handlers/base.py
shmufeld: Good catch. Implemented in 0e91a53 and hardened in 5403e5e: local filesystem and SQLite failures remain in MirrorResult.failures, increment the classified local_io counter, and no longer set backend_reachable=0. Only backend_failures affect backend reachability. One deliberate difference from the suggestion: local failures still make the initial reconcile unsuccessful and block readiness, preserving the HLD fail-closed startup contract. Strict presence and traversal checks ensure those failures are not silently converted into no-ops. Python 3.12 CI is green with 217 tests.

ufm-state-mirror/state_mirror/mirror.py
shmufeld: Yes—I’m keeping these references. They document the non-obvious backend-wins and SQLite deletion-onset policies and make the implementation traceable to the HLD requirements.

ufm-state-mirror/state_mirror/mirror.py
shmufeld: Agreed—this is a valid follow-up and is out of scope for this PR. I would avoid backend I/O directly inside the signal handler; the safer design is for SIGTERM to set a shutdown event and let the normal loop perform a bounded final drain and scan before stopping the observer, so termination cannot deadlock or hang indefinitely on a backend outage.

ufm-state-mirror/state_mirror/mirror.py
shmufeld: Implemented for drain_once() as well. Runtime local read failures are counted and retried without overriding the last known backend reachability; real backend failures still win the aggregate batch state.

ufm-state-mirror/state_mirror/mirror.py
shmufeld: I implemented the backend-probe part but intentionally kept result.succeeded as the readiness gate. Per the HLD and the locked startup policy, the initial reconcile must complete with zero failures; an unreadable required file means mirroring is incomplete, so readiness must remain false. A local-only zero-op scan now still probes the backend, allowing the accurate state backend_reachable=1, initial_reconcile_ok=0, and /ready=503.

ufm-state-mirror/state_mirror/mirror.py
shmufeld: Implemented consistently in poll_sqlite(). SQLite and local I/O errors remain reconcile failures and are classified as local_io, while only backend failures mark the backend unreachable.

ufm-state-mirror/state_mirror/handlers/directory.py
```diff
@@ -35,10 +36,14 @@ def _key_for_rel(self, relpath: str) -> str:
 
     def _iter_local_files(self):
         root = self.entry.path
-        if not os.path.isdir(root):
+        try:
```
ananalaghbar, line 39: can we reuse path_exists that you added in base.by?

ufm-state-mirror/state_mirror/mirror.py
```diff
@@ -51,6 +53,8 @@ def _reason(exc: BaseException) -> str:
     present (set at the store boundary, by the active backend's classifier),
     else fall back to a generic transport/OS classification.
     """
+    if isinstance(exc, sqlite3.Error):
+        return "local_io"
```
ananalaghbar: please define it as a constant for reuse it

ufm-state-mirror/state_mirror/handlers/directory.py
shmufeld: I considered reusing `path_exists()`, but kept this code as-is intentionally. `_iter_local_files()` needs the `stat_result` immediately afterward to validate `S_ISDIR`; calling `path_exists(root)` first would perform `os.stat()` twice and add another race window without removing any logic. The current single `os.stat()` has the same absence/error semantics: only `FileNotFoundError` becomes a no-op, while permission and I/O errors propagate. A shared helper would only help if we changed it to return `stat_result | None`, which feels unnecessary for this PR.

ufm-state-mirror/state_mirror/mirror.py
shmufeld: Agreed. Implemented in `dcf7b79`: `LOCAL_IO_REASON` is now defined once in `redis_errors.py` and reused by the canonical reason tuple, OSError classification, SQLite classification, and backend-failure filtering. Focused Ruff checks and `tests/test_redis_errors.py` pass (19 tests).

ananalaghbar (approved):

vg12345 (approved):

alextabachnik (approved):

PR #382, opened 2026-07-06
issue:5134987 [BugFix] [2/3] Harden state mirror path safety

shmufeld:
## What

Harden directory mirroring/restoration and classifier ownership validation.

- Reject generated `:meta` key collisions.
- Reject recursive and non-recursive filesystem ownership overlaps.
- Resolve classifier path aliases before overlap checks.
- Refuse symlinks in every directory path component.
- Use descriptor-relative traversal and atomic restore writes.
- Preserve prior identity best-effort and retain readable defaults for new children.
- Validate every backend child path before writing any entry.

## Why ?

Redmine [#5134987](https://redmine.mellanox.com/issues/5134987), under story [#5079041](https://redmine.mellanox.com/issues/5079041).

Lexical containment and final-component no-follow checks leave intermediate-directory race windows, while overlapping classifier ownership can mirror or restore the same file through multiple handlers.

## How ?

All directory reads and writes traverse from an opened root directory descriptor with no-follow checks for every component. Restore creates and replaces children relative to retained directory descriptors. Classifier validation compares resolved filesystem ownership and physical backend keyspaces.

This is **2/3** in a stacked series:

1. [PR #381](https://github.com/Mellanox/ufm_sdk_3.0/pull/381) — required base.
2. This PR — directory and classifier path safety.
3. [PR #383](https://github.com/Mellanox/ufm_sdk_3.0/pull/383) — SQLite replacement detection and logging parity.

## Testing ?

- Python 3.12: `ruff check .` — passed.
- Python 3.12: `ruff format --check .` — passed.
- Python 3.12: `python -m pytest -q` — **228 passed**.
- NVIDIA copyright validation — passed for the four-file delta.
- Fresh-context review — no actionable correctness regressions.
- GitHub Actions Docker build — passed.

### DIFF
--- a/ufm-state-mirror/state_mirror/classifier.py
+++ b/ufm-state-mirror/state_mirror/classifier.py
@@ -24,6 +24,7 @@ descriptive fields without breaking the load.
 from __future__ import annotations
 
 import logging
+import os
 from dataclasses import dataclass, field
 from enum import Enum
 from typing import Any, Optional
@@ -167,7 +168,7 @@ class Classifier:
         return classifier
 
     def _check_uniqueness(self) -> None:
-        """Reject duplicate paths and duplicate/overlapping keyspaces.
+        """Reject overlapping filesystem ownership and backend keyspaces.
 
         Reports *every* problem in one error rather than failing on the first,
         so a misconfigured classifier surfaces all its issues at once -- but
@@ -175,15 +176,25 @@ class Classifier:
         duplicates this also rejects *nested* keyspaces: a non-directory
         ``redis_key`` that falls under a directory ``redis_key_prefix`` (or one
         prefix nested under another) would collide on restore, where the
-        directory handler enumerates every key under its prefix.
+        directory handler enumerates every key under its prefix. A file nested
+        below a recursive directory is likewise rejected because two handlers
+        would otherwise own the same filesystem path.
         """
         errors: list[str] = []
         paths: set[str] = set()
         seen: list[tuple[str, bool]] = []  # (key_or_prefix, is_prefix), in order
+        entries_seen: list[Entry] = []
         for e in self.entries:
-            if e.path in paths:
+            normalized_path = os.path.realpath(e.path)
+            if normalized_path in paths:
                 errors.append(f"duplicate path: {e.path}")
-            paths.add(e.path)
+            paths.add(normalized_path)
+            for other in entries_seen:
+                if _filesystem_path_overlap(e, other):
+                    errors.append(
+                        f"filesystem paths {e.path!r} and {other.path!r} have overlapping ownership"
+                    )
+            entries_seen.append(e)
             key = e.redis_key_prefix if e.is_directory else e.redis_key
             if key is None:
                 continue  # missing key already reported by Entry.validate()
@@ -191,6 +202,11 @@ class Classifier:
                 if _keys_collide(key, e.is_directory, other_key, other_is_prefix):
                     errors.append(
                         f"redis key/prefix {key!r} ({e.path}) collides with {other_key!r}"
+                    )
+                elif _metadata_keys_collide(key, e.is_directory, other_key, other_is_prefix):
+                    errors.append(
+                        f"redis key/prefix {key!r} ({e.path}) collides with a metadata key "
+                        f"generated for {other_key!r}"
                     )
             seen.append((key, e.is_directory))
         if errors:
@@ -250,3 +266,35 @@ def _keys_collide(key_a: str, a_is_prefix: bool, key_b: str, b_is_prefix: bool) 
         or (a_is_prefix and key_b.startswith(key_a))
         or (b_is_prefix and key_a.startswith(key_b))
     )
+
+
+def _metadata_keys_collide(key_a: str, a_is_prefix: bool, key_b: str, b_is_prefix: bool) -> bool:
+    """Whether either entry can overwrite the other's generated ``:meta`` key."""
+    if not a_is_prefix and not b_is_prefix:
+        return key_a == key_b + ":meta" or key_b == key_a + ":meta"
+    if a_is_prefix != b_is_prefix:
+        prefix = key_a if a_is_prefix else key_b
+        exact_key = key_b if a_is_prefix else key_a
+        return (exact_key + ":meta").startswith(prefix)
+    return False
+
+
+def _filesystem_path_overlap(a: Entry, b: Entry) -> bool:
+    """Whether two handlers can mirror the same filesystem object."""
+    a_path = os.path.realpath(a.path)
+    b_path = os.path.realpath(b.path)
+
+    def contains(directory: str, candidate: str) -> bool:
+        try:
+            return os.path.commonpath([directory, candidate]) == directory
+        except ValueError:
+            return False
+
+    def owns(entry: Entry, directory: str, other: Entry, candidate: str) -> bool:
+        if not entry.is_directory:
+            return False
+        if entry.recursive:
+            return contains(directory, candidate)
+        return not other.is_directory and os.path.dirname(candidate) == directory
+
+    return owns(a, a_path, b, b_path) or owns(b, b_path, a, a_path)
--- a/ufm-state-mirror/state_mirror/classifier.py
+++ b/ufm-state-mirror/state_mirror/classifier.py
@@ -31,6 +31,8 @@ from typing import Any, Optional
 
 import yaml
 
+from state_mirror.wire import META_SUFFIX
+
 log = logging.getLogger(__name__)
 
 
@@ -269,13 +271,13 @@ def _keys_collide(key_a: str, a_is_prefix: bool, key_b: str, b_is_prefix: bool) 
 
 
 def _metadata_keys_collide(key_a: str, a_is_prefix: bool, key_b: str, b_is_prefix: bool) -> bool:
-    """Whether either entry can overwrite the other's generated ``:meta`` key."""
+    """Whether either entry can overwrite the other's generated metadata key."""
     if not a_is_prefix and not b_is_prefix:
-        return key_a == key_b + ":meta" or key_b == key_a + ":meta"
+        return key_a == key_b + META_SUFFIX or key_b == key_a + META_SUFFIX
     if a_is_prefix != b_is_prefix:
         prefix = key_a if a_is_prefix else key_b
         exact_key = key_b if a_is_prefix else key_a
-        return (exact_key + ":meta").startswith(prefix)
+        return (exact_key + META_SUFFIX).startswith(prefix)
     return False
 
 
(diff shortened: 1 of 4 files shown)


ufm-state-mirror/state_mirror/handlers/directory.py
ananalaghbar, line 50: From Cursor:

The two traversal branches now treat a symlinked child file differently:

- Non-recursive (line 57) uses de.is_file(follow_symlinks=False), so a symlink child is silently skipped (LOCAL_NOOP, no failure) — this is what test_directory_refuses_symlink_child asserts.
- Recursive (lines 47–53) yields symlink files from os.walk (a symlink-to-file lands in files), and _read_file_nofollow then raises OSError (ELOOP) → recorded as a failure on every scan/poll.
After PR #381 this is no longer a readiness blocker (the `OSError` is `local_io`, excluded from `backend_failures`), but a single benign symlink in a recursive tree — e.g. under a conf/plugins subtree — will emit a recurring mirror: child failed … log and bump backend_errors_total{reason="local_io"} forever, which can trip error-rate alerts. Please make the behavior consistent (recommend skipping symlinks in both branches).

```python
        if self.entry.recursive:
            for dirpath, _dirs, files in os.walk(root, onerror=self._raise_walk_error):
                for name in sorted(files):
                    full = os.path.join(dirpath, name)
                    # Match the non-recursive branch: a symlink child is not
                    # mirrored. Skipping here (instead of letting the O_NOFOLLOW
                    # open raise) avoids a recurring local_io failure/log on
                    # every scan for a benign symlink in the tree.
                    if os.path.islink(full):
                        continue
                    yield os.path.relpath(full, root), full
```

(This does reintroduce one `lstat` per child, which the current comment deliberately avoids — worth confirming the author prefers "skip" over "fail loud". If "fail loud" is intended, then the non-recursive branch should be aligned instead.)

ufm-state-mirror/state_mirror/handlers/directory.py
```diff
@@ -71,24 +70,95 @@ def _iter_redis_relpaths(self):
...
+    def _validate_relpath(relpath: str) -> list[str]:
+        parts = relpath.split(os.sep)
+        if not relpath or os.path.isabs(relpath) or any(part in ("", ".", "..") for part in parts):
+            raise wire.WireError(f"directory child has out-of-root path {relpath!r}")
+        return parts
+
+    def _open_parent(self, relpath: str, *, create: bool) -> tuple[int, str]:
+        """Open a child's parent via no-follow directory descriptors."""
+        parts = self._validate_relpath(relpath)
+        flags = os.O_RDONLY | getattr(os, "O_DIRECTORY", 0) | getattr(os, "O_NOFOLLOW", 0)
```
ananalaghbar, line 102: parent_fd = os.open(self.entry.path, flags) includes O_NOFOLLOW, so if the final component of a configured root is a symlink, both mirror() and restore() fail closed for that whole entry. That's a behavior change from the previous os.path.realpath(self.entry.path) approach. The root is operator-configured (trusted), so refusing to follow it may be stricter than intended — the danger is symlinks below the root. Can you confirm none of the roots in `state_mirror.yaml` resolve through a symlink? If any might, consider following the root but keeping O_NOFOLLOW on descendants:

```python
        flags = os.O_RDONLY | getattr(os, "O_DIRECTORY", 0) | getattr(os, "O_NOFOLLOW", 0)
        # The configured root is operator-trusted; follow it even if it is a
        # symlink. O_NOFOLLOW is enforced only on the descendant components.
        root_flags = os.O_RDONLY | getattr(os, "O_DIRECTORY", 0)
        parent_fd = os.open(self.entry.path, root_flags)
        try:
            for part in parts[:-1]:
                if create:
                    with suppress(FileExistsError):
                        os.mkdir(part, dir_fd=parent_fd)
                next_fd = os.open(part, flags, dir_fd=parent_fd)
```

ufm-state-mirror/state_mirror/handlers/directory.py
```diff
@@ -71,24 +70,95 @@ def _iter_redis_relpaths(self):
...
+        tmp = name + ".statemirror.tmp"
+        prev = None
+        try:
+            try:
+                prev = os.stat(name, dir_fd=parent_fd, follow_symlinks=False)
+                if not stat.S_ISREG(prev.st_mode):
+                    raise OSError(f"refusing to replace non-regular file: {relpath}")
+            except FileNotFoundError:
+                pass
+            flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC | getattr(os, "O_NOFOLLOW", 0)
```
ananalaghbar, line 142: Permission preservation only runs when `prev `is not `None` (an existing file, lines 139–155). A brand-new restored child is created at `0o666 & ~umask` (line 133). If the restore init container runs with a restrictive umask, or the main UFM container runs as a different UID than the sidecar, freshly-restored config files may not be readable where they need to be. `test_directory_restore_new_child_is_readable` (line 387: assert mode & 0o044) only passes because the test host umask is ≤ 022. Consider setting an explicit, deterministic mode for new children rather than relying on umask, and confirm the sidecar/main UID relationship makes 0o644 sufficient.

ufm-state-mirror/state_mirror/classifier.py
```diff
@@ -250,3 +266,35 @@ def _keys_collide(key_a: str, a_is_prefix: bool, key_b: str, b_is_prefix: bool)
         or (a_is_prefix and key_b.startswith(key_a))
         or (b_is_prefix and key_a.startswith(key_b))
     )
+
+
+def _metadata_keys_collide(key_a: str, a_is_prefix: bool, key_b: str, b_is_prefix: bool) -> bool:
+    """Whether either entry can overwrite the other's generated ``:meta`` key."""
+    if not a_is_prefix and not b_is_prefix:
+        return key_a == key_b + ":meta" or key_b == key_a + ":meta"
```
ananalaghbar: Please consider defining `:meta` in a constant and reuse it

ufm-state-mirror/state_mirror/classifier.py
```diff
@@ -167,23 +168,33 @@ def load_file(yaml_path: str) -> "Classifier":
...
+        below a recursive directory is likewise rejected because two handlers
+        would otherwise own the same filesystem path.
         """
         errors: list[str] = []
         paths: set[str] = set()
         seen: list[tuple[str, bool]] = []  # (key_or_prefix, is_prefix), in order
+        entries_seen: list[Entry] = []
         for e in self.entries:
-            if e.path in paths:
+            normalized_path = os.path.realpath(e.path)
```
ananalaghbar, line 190: NIT:
When two entries resolve to the same realpath, you now get both duplicate path: … (line 190) and filesystem paths … have overlapping ownership (line 195), because contains() returns True for directory == candidate. Not wrong, just noisy. Optional: short-circuit _filesystem_path_overlap when a_path == b_path so the duplicate-path error stands alone.

ananalaghbar (dismissed):

ufm-state-mirror/state_mirror/handlers/directory.py
shmufeld: Done in `03587af`: recursive traversal now skips symlink file children the same way as the non-recursive branch, avoiding recurring `local_io` logs/metrics for benign symlink children. I also added `test_recursive_directory_skips_symlink_child` to lock the behavior. Symlink parent components still fail closed through the descriptor-relative no-follow open path.

ufm-state-mirror/state_mirror/handlers/directory.py
shmufeld: I kept this one unchanged intentionally. The security boundary in this PR is to refuse symlinks in every directory path component, including the configured root final component. Following the root would allow that root component to be replaced by a symlink and still be accepted before descendant `O_NOFOLLOW` checks start.

The bundled classifier example uses direct `/opt/ufm/files/...` roots. The authoritative consumer `state_mirror.yaml` should still be checked to confirm it does not rely on symlink roots; if it does, I think the safer fix is to make the configured root a real directory rather than relaxing the engine boundary.

ufm-state-mirror/state_mirror/handlers/directory.py
shmufeld: Done in `03587af`: new restored children are now explicitly `fchmod`d to `0644` before the atomic replace, so readability no longer depends on the restore process umask. I also tightened the test to run under `umask 077` and assert the deterministic mode.

ufm-state-mirror/state_mirror/classifier.py
shmufeld: Done in `03587af`: classifier metadata collision logic now reuses `wire.META_SUFFIX` instead of hardcoding `:meta`.

ufm-state-mirror/state_mirror/classifier.py
shmufeld: Done in `03587af`: `_filesystem_path_overlap` now short-circuits when the resolved realpaths are equal, so duplicate paths only report the duplicate-path error. Added a focused test for that case.

alextabachnik (dismissed):

ufm-state-mirror/state_mirror/handlers/directory.py
```diff
@@ -71,24 +71,101 @@ def _iter_redis_relpaths(self):
...
+    def _open_parent(self, relpath: str, *, create: bool) -> tuple[int, str]:
+        """Open a child's parent via no-follow directory descriptors."""
+        parts = self._validate_relpath(relpath)
+        flags = os.O_RDONLY | getattr(os, "O_DIRECTORY", 0) | getattr(os, "O_NOFOLLOW", 0)
+        parent_fd = os.open(self.entry.path, flags)
+        try:
+            for part in parts[:-1]:
+                if create:
+                    with suppress(FileExistsError):
+                        os.mkdir(part, dir_fd=parent_fd)
```
vg12345: New files get a deterministic `0o644`, but intermediate dirs created here use `0o777 & ~umask`, so under a restrictive umask they end up `0o700`. If the main UFM container runs as a different UID, it can't traverse restored subdirs (e.g. `plugins/tools/nvp`). Could we set a deterministic mode on created dirs too, like we do for files?

vg12345 (dismissed):

ufm-state-mirror/state_mirror/handlers/directory.py
shmufeld: Done in `74a964b`: newly created intermediate restore directories now use `DEFAULT_RESTORED_DIR_MODE = 0o755` and are `fchmod`d after the descriptor-relative open, so a restrictive `umask` cannot leave them at `0700`. Existing directories keep their current mode, and restored new files still use deterministic `0644`. Added a regression test under `umask 077` for `plugins/preserved/tools/nvp/new.conf`. Validation passed: `pytest -q` -> 231 passed, `ruff check .`, `ruff format --check .`, `git diff --check`, copyright check, and fresh pre-push review (`APPROVE`).

ananalaghbar (approved):

vg12345 (approved):

alextabachnik (approved):

PR #387, opened 2026-07-16
issue:5158735 [Build] automate StateMirror release artifact

shmufeld:
## What
Automate the `ufm-state-mirror` sidecar release path through the existing `UFM_PLUGINS_SDK_RELEASE` Blossom job.

This PR:
- Adds `ufm-state-mirror` to the stable and nbuprod release dropdowns.
- Adds StateMirror-owned stable and nbuprod release matrices under `ufm-state-mirror/.ci`, without changing the shared plugin release matrices.
- Runs the release build on the Blossom `swx-k8s-spray` Kubernetes worker with the Docker socket and `/auto/mswg` mounted.
- Publishes one immutable artifact per version under `/auto/mswg/release/ufm/ufm-state-mirror/<VERSION>/`.
- Maintains a relative `latest` symlink to the newest successfully published artifact.
- Serializes stable, nbuprod, and manual releases and uses staged atomic publication with interruption recovery.
- Requires numeric `MAJOR.MINOR.PATCH`, rejects duplicate versions and `latest` downgrades, and validates existing link targets.
- Bumps StateMirror to `1.0.1` in both `VERSION` and `state_mirror.__version__`.
- Updates `ufm-state-mirror/BUILD.md` with the automated Blossom release flow and manual fallback.

## Why ?
Redmine: https://redmine.mellanox.com/issues/5158735
Jira context: SWXUFM-623

The StateMirror sidecar image release was still manual even though the repository already has Blossom release automation for Docker artifacts. This adds an automated release path while keeping StateMirror as a standalone top-level component rather than moving it under `plugins/`.

Versioned, immutable artifacts let UFM packaging discover the newest StateMirror through `latest` while pinning each versioned UFM package to a concrete StateMirror artifact.

## How ?
The stable and nbuprod jobs select the StateMirror-specific matrix through the `conf_file` parameter. Existing plugins continue using the shared matrices unchanged.

Both StateMirror matrices dispatch to `ufm-state-mirror/.ci/release_build.sh`. The helper:

- Verifies `PLUGIN_VERSION` against `ufm-state-mirror/VERSION`.
- Uses one release lock across all publication paths.
- Builds and validates the gzip artifact in a hidden staging directory.
- Atomically renames the staging directory to the immutable version directory.
- Atomically replaces `latest` with a relative symlink after successful publication.
- Rolls back ordinary publication failures and safely resumes a hard-interrupted pending publication.
- Validates the current `latest` target and refuses to move it to an older numeric version.

Expected layout:

```text
[...]

PR #388, opened 2026-07-30
issue:5158735 [Build] fix StateMirror nbuprod release flow

shmufeld:
## What
Fix the StateMirror nbuprod release flow: run it in Michael Privorotsky's approved x86_64 DIND builder as a privileged Blossom Kubernetes pod, publish readable immutable artifacts, make legacy `latest` links self-healing, and remove the unsupported StateMirror GitHub Actions workflow. Release validation build `1.0.1-8`, grouped with other `1.0.1` artifacts.

## Why ?
Follow-up for [issue 5158735](https://redmine.mellanox.com/issues/5158735). Nbuprod validation exposed an incorrect Kubernetes cloud, the `/auto/mswg` to `/auto/sw` filesystem alias, restrictive staging permissions, an already-published `1.0.1`, and legacy link/rollback state that prevented a new release.

Michael provided matching x86_64 and aarch64 DIND builders. StateMirror publishes one architecture-neutral immutable artifact per version, so this matrix uses only the x86_64 builder and avoids two architectures racing to publish the same path.

## How ?
Use `nbu-harbor.gtm.nvidia.com/ufm/ci/dind_ufm_x86_64:20251224` as an explicitly selected ci-demo `tool` container in Blossom cloud `il-ipp-blossom-prod`, namespace `swx`. Run the pod privileged, start its nested Docker daemon, and wait for readiness. Mount the logical `/auto/mswg` release path and narrowly scoped `/auto/sw/release/ufm` physical alias; no bare-metal node or host Docker socket is used.

Remove `.github/workflows/state-mirror-ci.yml` because GitHub Actions are not an approved CI method for this group. StateMirror validation remains in the approved Jenkins/ci-demo release flow and documented local test commands.

The committed `VERSION` file declares `BASE_VERSION=1.0.1` and `BUILD_NUMBER=8`. The release helper validates both fields, derives `EXTENDED_VERSION=1.0.1-8`, and requires Jenkins `PLUGIN_VERSION` to match it. Bumping the base or build therefore changes only one source field.

Group build artifacts directly under their base version:

```text
1.0.1/
├── ufm-state-mirror_1.0.1-docker.img.gz
├── ufm-state-mirror_1.0.1-8-docker.img.gz
└── ufm-state-mirror_1.0.1-9-docker.img.gz
```

Each extended artifact remains immutable. Publication stages only the new artifact, preserves other images in the base directory, and uses a build-specific pending marker for interruption recovery. Numeric downgrade protection covers both base versions and build suffixes.

Recognize canonical grouped, physical-alias, relative, and legacy flat `latest` targets. Both release matrices pass the `/auto/sw` alias explicitly so recognition does not depend on Kubernetes bind-mount path resolution. Preserve downgrade validation, then replace a recognized legacy or stale target only after the new artifact succeeds. Equality never authorizes rebuilding an immutable version without a validated pending marker. Remove legacy `.latest.rollback.*` residue only when it has the expected safe structure; unexpected data is preserved and fails the release.

Published artifacts use `0644`; base-version directories use group-writable setgid mode `2775` with `sw_ufm` GID 4200. Both release pods run with GID 4200. The helper verifies the final GID/mode and performs a temporary write probe before Docker work, so permission drift fails immediately with ownership details. In nbuprod, the nested Docker daemon remains root, while the Docker client and release helper run as numeric `swx-jenkins` UID 6213/GID 4200. The ephemeral Docker socket is restricted to that group and checked before the build. This avoids the release export's root-squash behavior without weakening artifact permissions. Release locking is temporarily disabled to bypass the stale lock owned by the former publisher identity; release jobs must not overlap until locking is restored, and legacy rollback-residue cleanup currently runs without lock protection. After successful artifact publication, `latest` is updated with `unlink` and `ln -s` as requested for the non-overlapping UFM deployment flow. Immutability, validation, and pending-marker recovery remain intact.

## Testing ?
- `ufm-state-mirror/.ci/test_release_build.sh` — passed on a prior Linux validation run; it covers grouped multi-build publication, preservation of the existing base artifact, duplicate and equal-dangling rejection, interrupted-publication resume, injected post-symlink SIGTERM recovery, explicit bind-mount alias recognition, relative/physical/legacy-flat migration, stale-link recovery, downgrade rejection, rollback-residue handling, Blossom-compatible chmod syntax, exact GID/mode postconditions, write-probe failure before Docker, wrong-GID rejection, and invalid version metadata.
- `bash -n` for the release helper, integration test, and Docker build helper — passed.
- `python -m pytest -q` — 237 passed.
- `ruff check .` — passed.
- `ruff format --check .` — passed.
- ci-demo schema validation for both StateMirror release matrices — passed.
- Python copyright check — passed; no Python files changed.
- PR title validation — passed.
- Final fresh-context review includes the authorized temporary lock removal and unlink/symlink `latest` update — APPROVE with no Critical, Important, or Minor findings.
- StateMirror GitHub Actions workflow removed per group policy; no GitHub Actions coverage remains for this component.
- Nbuprod build #59 reported success but did not execute the `Build StateMirror` step; it only started and removed a container on `blue-force01`, so it did not validate or publish an artifact.
- Nbuprod build #67 validated the Blossom pod, nested Docker, mounts, full image build, compressed artifact, and integrity check. Its write failure identified the shared GID requirement now addressed by `runAsGroup: 4200`, directory mode `2775`, and the pre-build write probe; `1.0.1-8` remains the validation build.
- Nbuprod build #68 exposed Jenkins' Java `String` requirement for `runAsGroup`; both matrices now pass quoted `"4200"`, matching ci-demo's advanced example and the plugin setter.
- Nbuprod build #69 proved that GID 4200 and directory mode `2775` alone are insufficient: the NFS export root-squashes UID 0. The nbuprod step now retains root only for `dockerd` and runs Docker publication as UID 6213/GID 4200, with identity and Docker-access preflights.
- Nbuprod build #71 showed that this Alpine image's `sudo` cannot resolve an unregistered numeric GID. The step now resolves or creates `sw_ufm` as GID 4200, adds `swx-jenkins` to it for the image's `root ALL=(ALL)` sudo policy, and invokes `sudo` with resolvable names.
- Nbuprod build #73 validated UID/GID and Docker access, then failed on the stale `.release.lock` created by the former publisher identity. Locking is temporarily disabled for a single-job validation run and must be restored before concurrent release execution is allowed.
- Nbuprod build #74 completed successfully and published `1.0.1/ufm-state-mirror_1.0.1-8-docker.img.gz`; `latest` now points to that artifact.

## Special triggers

* `bot:retest` rerun Jenkins CI (to rerun GitHub CI, use the **Checks** tab and rerun all jobs)
* `bot:upgrade` run additional update tests


ananalaghbar (dismissed):

ufm-state-mirror/.ci/matrix_job_release_nbuprod.yaml
```diff
@@ -6,23 +6,16 @@ registry_path: /swx-storage/ci-demo
...
   - {credentialsId: '425bb907-c357-4fde-92e0-67854a857b4f', usernameVariable: 'UFM_USER', passwordVariable: 'UFM_PASS'}
 
-volumes:
-  - {mountPath: /var/run/docker.sock, hostPath: /var/run/docker.sock}
-  - {mountPath: /auto/mswg, hostPath: /auto/mswg}
-
-runs_on_dockers:
-  - {file: '.ci/Dockerfile', arch: 'x86_64', name: 'plugin_worker', tag: 'latest'}
+runs_on_agents:
+  - nodeLabel: 'SWX-CI-DOCKER'
```
vg12345: This puts nbuprod back on `SWX-CI-DOCKER`. In #387 @mprivorotsky asked to move off it, because that label points to blueforce01, outside UFM. Can you get his OK here? Please also note in the file that stable stays on k8s, and why nbuprod needs the agent (its executor does not support the k8s/container flow).

shmufeld: right.
it is not working on other agents.
@mprivorotsky please approve this PR

vg12345 (dismissed):

ufm-state-mirror/.ci/matrix_job_release_nbuprod.yaml
```diff
@@ -6,23 +6,21 @@ registry_path: /swx-storage/ci-demo
...
+  - {mountPath: /auto/sw/release/ufm, hostPath: /auto/sw/release/ufm}
 
 runs_on_dockers:
-  - {file: '.ci/Dockerfile', arch: 'x86_64', name: 'plugin_worker', tag: 'latest'}
+  - {url: 'nbu-harbor.gtm.nvidia.com/ufm/ci/dind_ufm_x86_64:20251224', name: 'docker_builder', arch: 'x86_64', category: 'tool', nodeLabel: 'SWX-CI-DOCKER'}
 
 steps:
   - name: Build StateMirror
-    containerSelector: "{name: 'plugin_worker'}"
+    agentSelector: "{nodeLabel: 'SWX-CI-DOCKER'}"
```
mprivorotsky: you dont need to mention a node now

mprivorotsky (dismissed):

ufm-state-mirror/.ci/matrix_job_release_nbuprod.yaml
shmufeld: fixed

ananalaghbar (dismissed):

ufm-state-mirror/.ci/matrix_job_release_nbuprod.yaml
```diff
@@ -31,6 +33,18 @@ steps:
...
+        sleep 2
+      done
+      if ! docker info >/dev/null 2>&1; then
+        echo -e "Error: Docker did not become available within 120 seconds."
+        cat /var/log/docker.log
+        exit 1
+      fi
+
       RELEASE_ROOT="/auto/mswg/release/ufm/ufm-state-mirror"
       bash -x ufm-state-mirror/.ci/release_build.sh "${PLUGIN_VERSION}" "${RELEASE_ROOT}"
```
shahar-spormas: `PLUGIN_VERSION` here is `1.0.1`, and `1.0.1` is already published on the share (Jul 20). So this line always dies at `A build for ufm-state-mirror 1.0.1 already exists`, which is where #62 and #63 stopped.

That guard runs *before* everything this PR adds: the `chmod` calls and the absolute `latest` write. Those builds proved the DIND worker, the mounts and the lock, but the changed code has never actually executed.

This is also one of the reasons we need a patch/build version such as `1.0.1-8`, so a rebuild of the same source version can still publish. Note the script would reject that today: the format check only accepts numeric `MAJOR.MINOR.PATCH`, and the `latest` target name is rebuilt from that same string, so both would need to allow the suffix.

For now, please bump `VERSION` and run once so the publish path completes for real.

ufm-state-mirror/.ci/release_build.sh
```diff
@@ -153,20 +162,30 @@ fi
 
 if [ -L "${LATEST_LINK}" ]; then
     CURRENT_TARGET="$(readlink "${LATEST_LINK}")"
-    CURRENT_VERSION="${CURRENT_TARGET%%/*}"
+    case "${CURRENT_TARGET}" in
+        "${RELEASE_ROOT}/"*)
+            CURRENT_RELATIVE_TARGET="${CURRENT_TARGET#"${RELEASE_ROOT}/"}"
+            ;;
+        *)
+            echo -e "Error: ${LATEST_LINK} must contain an absolute target under the release root."
```
shahar-spormas: The old `latest` symlink is never unlinked. If it exists and isn't exactly what this check expects, the script just `exit 1`s, and the release stays blocked until someone unlinks it by hand on `/auto/mswg`.

Cases that hit it: a dangling link, the old relative target (`1.0.1/...`), or the `/auto/sw` spelling — the last one newly reachable now that the pod mounts both paths.

It should `unlink` the old symlink first and then create the new one, instead of bailing out. The downgrade guard should stay strict.

ufm-state-mirror/.ci/release_build.sh
```diff
@@ -195,12 +214,19 @@ if [ "${RESUME_PUBLISH}" = false ]; then
...
     STAGING_DIR=""
     VERSION_PUBLISHED=true
 fi
 
+if [ "${RESUME_PUBLISH}" = true ]; then
+    chmod 0644 "${VERSION_DIR}/${ARTIFACT_NAME}"
+    chmod u+rwx,go+rx "${VERSION_DIR}"
+fi
+
 LATEST_TMP_DIR="$(mktemp -d "${RELEASE_ROOT}/.latest.staging.XXXXXX")"
```
shahar-spormas: There is a leftover rollback link on the share that needs cleaning: `.latest.rollback.vxbY8y`, mode `700`, owned by sfeld.

It is residue from the earlier rollback logic — this version of the script does not even create that name — and nothing removes it.

ufm-state-mirror/.ci/matrix_job_release_nbuprod.yaml
shmufeld: I know.
I will bump the version in a new branch.
Didn't want to do it here

shahar-spormas:
1. dont add github action, this method is not supported at our group and not approved.
2. 
image as we can see it runs everything twice, I think on push is enough. (Either way, we don't want to use the GitHub Actions.)

shahar-spormas (approved):

mprivorotsky (approved):

vg12345 (approved):

